PR #29, opened 2018-05-08
Add function to reduce data in blocks

### LINKED ISSUE #24: Function apply a reduction operator in blocks
Create a `block_apply(easting, northing, data, reduction, spacing, region=None, adjust='spacing') -> easting, northing, data` function that takes a reduction function like `np.mean` and applies it to the data in windows. Returns the easting and northing coordinates of the centers of windows that have data and the reduction of the data in those windows. Region can be inferred from input and the spacing or region can be adjusted to fit each other like in `grid_coordinates`. 

This will serve as GMT's `blockmedian` and `blockmean` but can be more general.

leouieda:
Function `block_reduce` applies a reduction operation to data in blocks.
Also adds functions `get_region` to get the region from coordinates, 
`block_region` to generate the block boundaries (a generator), and
`inside` to get check which points are inside a given region. `inside`
allows passing in output and temporary arrays to avoid the overhead of
allocation in `block_reduce`.

Fixes #24

### DIFF
--- a/examples/block_reduce.py
+++ b/examples/block_reduce.py
@@ -1,0 +1,43 @@
+"""
+Data decimation in blocks
+=========================
+
+When gridding data that has been highly oversampled in a direction (shipborne
+and airborne data, for example), it is important to decimate the data before
+interpolation to avoid aliasing. Function :func:`verde.block_reduce` decimates
+data by applying a reduction operation (mean, median, mode, max, etc) to the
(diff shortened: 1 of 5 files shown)


leouieda:
pylint is complaining of too many local variables in block_reduce. Gotta find a way to make it simpler.

leouieda:
Maybe split the `for` loops that iterate over the blocks into a separate function. This might be useful in other contexts, like Euler deconv.

PR #67, opened 2018-05-27
Implement the Vector2D gridder

leouieda:
Implements the method of Sandwell and Wessel (2016) for gridding
2-component vector data using elastic Green's functions in 
`verde.Vector2D`.
Extends the method by using ridge regression instead of SVD
and adding the option of laying the forces out in a grid (which is 
required for KFold cross-validation because otherwise the model
would change for different data folds).

Refactors `verde.Spline` so that it can be used as a base class for
`verde.Vector2D` to avoid code repetition.


leouieda:
TODO:

- [ ] Write an example using the simple classic solution without regularization
- [ ] Split out the cross-validation stuff into a separate PR
- [ ] Write tests for Vector2D
- [ ] Write the docstrings for Vector2D

PR #92, opened 2018-07-25
Use pykdtree if available in blockreduce and distance_mask

### LINKED ISSUE #90: Use pykdtree in BlockReduce to speed up computations
## Feedback or description of feature requested

[pykdtree](https://github.com/storpipfugl/pykdtree) has the same API as `cKDTree` but @djhoese showed that it can be 2x faster on a `BlockReduce` example. We should add it as an optional dependency. Use it if we can import it or fallback to `cKDTree` if not. Should include some instructions on the `docs/install.rst` page and a mention in the `BlockReduce` docstring.

djhoese:
Adds optional 'pykdtree' dependency for blockreduce and grid_math modules for better performance on the associated operations. Include notes in the docstrings for these methods and in the install instructions.

Feel free to be picky about how things are worded 😉 I get it.

Fixes #90

### DIFF
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -25,6 +25,7 @@ Dependencies
 * `xarray <http://xarray.pydata.org/>`__
 * `scikit-learn <http://scikit-learn.org/>`__
 * `requests <http://docs.python-requests.org/>`__
+* `pykdtree <https://github.com/storpipfugl/pykdtree>`__ (optional)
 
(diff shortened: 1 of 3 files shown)


leouieda:
@djhoese look great to me! The failures on the CIs are my fault. I'll make quick fix and then we can merge this. Thanks!

leouieda:
OK, done! Thanks for implementing this. Merging :tada:

PR #139, opened 2018-10-10
Flake8 check, simplify Makefile and Travis builds

leouieda:
Add flake8 to the `make check` command to get a quick quality check
(pylint is slow). Separate pylint on `make lint` which can be run less
often.

Took the opportunity to rename a few make commands:
* `develop` is now `install`, which is what it actually does
* `test` and `coverage` are merged because it doesn't hurt to always
  calculate coverage. This avoids repeated code in the Makefile.

Replace the Python 3.5 OSX Travis build with a Linux build that only
runs the style checks. This will be fast and quickly alert to style
problems while the other builds continue.

**Reminders**

- [x] Add new public functions/methods/classes to `doc/api/index.rst`.
- [x] Write detailed docstrings for all functions/methods.


leouieda:
The appveyor failure is known problem with scikit-learn 0.20.0 (scikit-learn/scikit-learn#12226) will ignore until the bug fix release.

PR #148, opened 2018-10-18
Function to convert grids to tables

jessepisel:
takes an xarray grid and transforms it into a pandas dataframe. Still needs more tests, PR to see if it produces any errors. This adds functionality for issue #5 

Fixes #

**Reminders**

- [x] Add new public functions/methods/classes to `doc/api/index.rst`.
- [x] Write detailed docstrings for all functions/methods.


leouieda:
Thanks @jessepisel! Here are some ideas for tests:

* Create a small 2D `xarray.DataArray` with an `np.arange(20).reshape(shape)` as the data. Then compare it with the known output that you expect. This way we can test that the raveling of the 2D arrays is done correctly. This could be implemented as a doctest in the function docstring to serve as an example.
* Make a `xarray.Dataset` with a single `DataArray` and check that the output is correct.
* Make a `xarray.Dataset` with two `DataArrays` and check that the output is correct.

For all of these, use some tiny data that would be easy to check.

verde/utils.py
```diff
@@ -201,3 +203,49 @@ def maxabs(*args):
     arrays = [np.atleast_1d(i) for i in args]
     absolute = [np.abs([i.min(), i.max()]).max() for i in arrays]
     return np.max(absolute)
+
+
+def grid_to_table(grid):
+    """
+    Converts an xarray Dataset grid to a Pandas DataFrame
```
leouieda:
```suggestion
    Convert a grid to a table with the values and coordinates of each point.
```

verde/utils.py
```diff
@@ -201,3 +203,49 @@ def maxabs(*args):
...
+
```
leouieda: Please add a description of what the function does and what the user should expect from the output. It's good to mention things like: 1. the output coordinates are put through a meshgrid, 2. any restrictions (only works for 2d, or can be ndm, multiple data variables or only a single). 3. possible use cases for this function (to pass along grid data to functions that expect xyz, like `BlockReduce`).

verde/utils.py
```diff
@@ -201,3 +203,49 @@ def maxabs(*args):
...
+    Parameters
+    ----------
+    grid
```
leouieda:
```suggestion
    grid : :class:`xarray.Dataset`
```

This way sphinx will create a nice link to the xarray docs.

verde/utils.py
```diff
@@ -201,3 +203,49 @@ def maxabs(*args):
...
+        An xarray grid Dataset composed of X and Y coordinates and a variable.
+
+    Returns
+    -------
+    grid_to_table : DataFrame
```
leouieda:
```suggestion
    table : :class:`pandas.DataFrame`
```

verde/utils.py
```diff
@@ -201,3 +203,49 @@ def maxabs(*args):
...
+
+
+def grid_to_table(grid):
+    """
+    Converts an xarray Dataset grid to a Pandas DataFrame
+
+    Parameters
+    ----------
+    grid
+        An xarray grid Dataset composed of X and Y coordinates and a variable.
```
leouieda:
```suggestion
        A 2D grid with a single data variable.
```

The type is already specified in the line above.

verde/utils.py
```diff
@@ -201,3 +203,49 @@ def maxabs(*args):
...
+
+    Returns
+    -------
+    grid_to_table : DataFrame
+        Pandas DataFrame with coordinates and variables and increasing index
```
leouieda:
```suggestion
        :class:`pandas.DataFrame` with coordinates and variables.
```

verde/utils.py
```diff
@@ -201,3 +203,49 @@ def maxabs(*args):
...
+    ]  # extracts the variable name from the dataarray
+    coord_one = grid.coords[
+        coordinate_names[0]
+    ].values  # gets the first set of coordinates as a list
+    coord_two = grid.coords[
+        coordinate_names[1]
+    ].values  # gets the second set of coordinates as a list
+    locations = np.asarray(
+        list(itertools.product(coord_one, coord_two))
+    )  # creates a numpy array from the coordinates
```
leouieda: This could be replaced by `coordinates = [i.ravel() for i in np.meshgrid(coord_one, coord_two)]`. It's probably faster than `itertools.product` for large datasets but that probably doesn't matter.

verde/utils.py
```diff
@@ -201,3 +203,49 @@ def maxabs(*args):
...
+    variable = grid[variable_name[0]].values.flatten(
+        order="F"
+    )  # extracts the variable values for each location
+    data_dict = {
+        coordinate_names[0]: locations[:, 0],
+        coordinate_names[1]: locations[:, 1],
```
leouieda: If using `np.meshgrid`, these can become `coordinate_names[0]: coordinates0],`

verde/utils.py
```diff
@@ -201,3 +203,49 @@ def maxabs(*args):
...
+        Pandas DataFrame with coordinates and variables and increasing index
+
+    Examples
+    --------
+
+
+    """
+    coordinate_names = [
+        *grid.coords.keys()
+    ]  # extracts the coordinate names from the dataarray
```
leouieda: Kudos the wealth of comments in the code! Some of these can probably be removed because reading the code will say exactly the same as the comment. That is one of the nice things about Python. The comments can be reserved for less obvious bits of code, like the `itertools.product` below.

verde/utils.py
jessepisel: Updated pretty much everything. Thank you so much for the constructive feedback. Still having issues getting the documentation build to include the docstring.

leouieda: You'll need to:

1. Import your function on `verde/__init__.py` to make part of the API. This way it can be accessed as `verde.grid_to_table` and we can move it around overcas internally without breaking the API.
2. Include it in `doc/api/index.rst`. This should be enough for Sphinx to pick it up

jessepisel: I was missing the import. It looks like it built the documentation locally, I'm going to commit the changes.

leouieda: OK, the docs look good. Now it's a matter of getting the tests passing. I'll leave some comments below.

verde/utils.py
```diff
@@ -201,3 +202,93 @@ def maxabs(*args):
...
+        A 2D grid with a single data variable.
+
+    Returns
+    -------
+    table : :class:`pandas.DataFrame`
+        :class:`pandas.DataFrame` with coordinates and variables.
+
+    Examples
+    --------
+    >>> var1 = xr.DataArray(np.arange(20).reshape(4,5), coords = (np.arange(4), np.arange(5)))
```
leouieda: You'll to import xarray and numpy in the test:

```suggestion
    >>> import xarray as xr
    >>> import numpy as np
    >>> var1 = xr.DataArray(np.arange(20).reshape(4,5), coords = (np.arange(4), np.arange(5)))
```

verde/utils.py
```diff
@@ -201,3 +202,93 @@ def maxabs(*args):
...
+    grid : :class:`xarray.Dataset`
+        A 2D grid with a single data variable.
+
+    Returns
+    -------
+    table : :class:`pandas.DataFrame`
+        :class:`pandas.DataFrame` with coordinates and variables.
+
+    Examples
+    --------
```
leouieda: In general, it's best to use realist sounding variable names, like "easting" instead of "dim_0" or "temperature" instead of "var1". That helps keep things clear to the user, who will probably have variables like that in their code.

verde/utils.py
```diff
@@ -201,3 +202,93 @@ def maxabs(*args):
...
     return np.max(absolute)
+
+
+def grid_to_table(grid):
+    """
+    Convert a grid to a table with the values and coordinates of each point.
+
+    Takes a 2D grid as input, extracts the coordinates and runs them through
+    :func:`numpy.meshgrid` to create a 2D table. Works for 2D grids and
+    n-dimensional variables. Use cases includes passing gridded data to
```
leouieda:
```suggestion
    any number of variables. Use cases includes passing gridded data to
```

The "variables" are 2D but there can be any number of them.

verde/utils.py
```diff
@@ -201,3 +202,93 @@ def maxabs(*args):
...
+    coordinate_names = [*grid.coords.keys()]
+    coord_one = grid.coords[coordinate_names[0]].values
+    coord_two = grid.coords[coordinate_names[1]].values
+    coordinates = [i.ravel() for i in np.meshgrid(coord_one, coord_two)]
+    coord_dict = {
+        coordinate_names[0]: coordinates[0],
+        coordinate_names[1]: coordinates[1],
+    }
+    variable_name = [*grid.data_vars.keys()]
+    for i in enumerate(variable_name, 1):
```
leouieda: Is the `enumerate` needed? You can index the `grid` with the variable name and don't really need an index.

verde/utils.py
```diff
@@ -201,3 +202,93 @@ def maxabs(*args):
...
+    10      2      2    10    30
+    11      3      2    11    31
+    12      0      3    12    32
+    13      1      3    13    33
+    14      2      3    14    34
+    15      3      3    15    35
+    16      0      4    16    36
+    17      1      4    17    37
+    18      2      4    18    38
+    19      3      4    19    39
```
leouieda: There seems to be something wrong with the generation of the coordinates. The value at `[3, 0]` should be 15 from the `print(var1)`. From https://github.com/fatiando/verde/blob/master/verde/coordinates.py#L433 the coordinate order in the meshgrid should be `dim_1, dim_0` instead. This is actually a good example of why using more readable variable names in the tests makes it easier to reason about the code.

verde/utils.py
```diff
@@ -201,3 +202,96 @@ def maxabs(*args):
...
+    -------
+    table : :class:`pandas.DataFrame`
+        :class:`pandas.DataFrame` with coordinates and variables.
+
+    Examples
+    --------
+    >>> import xarray as xr
+    >>> import numpy as np
+    >>> temperature = xr.DataArray(np.arange(20).reshape(4,5),
+        coords = (np.arange(4), np.arange(5)), dims=['northing', 'easting'])
```
leouieda: @jessepisel you need to add `...` to get line continuations in doctests, like this:

```
>>> temperature = xr.DataArray(
...     np.arange(20).reshape(4,5),
...     coords=(np.arange(4), np.arange(5)), 
...     dims=['northing', 'easting']
... )
```

leouieda:
@jessepisel I pushed some changes to your branch to fix some formatting issues. The tests should pass now. As soon as the CIs are happy, I'll merge this in. Thanks for your help!

Quick tip: it's best to create a separate branch to work in and keep your `master` clean and updated with the main repository. After this is merged, your master branch will be out of sync with a bunch of commits that won't be in the main repo. You'll probably have to delete it to sync. Not doing that might cause issues for other PRs you open in the future. Let me know if you need help doing that.

jessepisel:
@leouieda Thank you for guiding me through all of this. I have learned a ton about formatting and docstrings, and tests over the past few weeks, thanks for letting me contribute and giving me a hand on my first open source contribution.

leouieda:
@jessepisel no problem! It's my pleasure.

These things can be tricky, specially the testing. I just pushed a new commit fixing the doctests in your function. They fail on some platforms because pandas prints the dataframes slightly differently. I've been bitten by this before but always end up making the same mistake. It's usually better to print the numpy arrays and try to use int types in doctests whenever possible.

leouieda:
@jessepisel congratulations on merging your first feature! :tada: 

Please feel free to open a new PR to add yourself to the `AUTHORS.md` file. Hope to get some more PRs from you :wink:

PR #160, opened 2018-12-14
Cast cross_val_score as a numpy.array

### LINKED ISSUE #155: Return an array from cross_val_score
Function `verde.model_selection.cross_val_score` fits gridders to many folds of the data and calculates the score for each fold. It currently returns a list of the scores, which is annoying if want to calculate the mean score. For example  `cross_val_score(...).mean()` doesn't work and needs to be `np.mean(cross_val_score(...))`. So it would be better if `cross_val_score` returned a numpy array instead.

Goto15:
In order to allow the use of mean(), cast the cross_val_score return data structure as a numpy.array. Also changed the in code comments to reflect the new return data structure. 

Fixes #155

### DIFF
--- a/verde/model_selection.py
+++ b/verde/model_selection.py
@@ -146,8 +146,8 @@ def cross_val_score(estimator, coordinates, data, weights=None, cv=None, client=
 
     Returns
     -------
-    scores : list
-        List of scores for each split of the cross-validation generator. If
+    scores : array
+        Array of scores for each split of the cross-validation generator. If
         *client* is not None, then the scores will be futures.
 
     Examples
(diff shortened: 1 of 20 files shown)


verde/model_selection.py
```diff
@@ -158,7 +158,7 @@ def cross_val_score(estimator, coordinates, data, weights=None, cv=None, client=
     >>> data = 10 - coords[0] + 0.5*coords[1]
     >>> # A linear trend should perfectly predict this data
     >>> scores = cross_val_score(Trend(degree=1), coords, data)
-    >>> print(', '.join(['{:.2f}'.format(score) for score in scores]))
+    >>> print('{:.2f}'.format(scores), sep = ", ")
```
leouieda:
```suggestion
    >>> print(', '.join(['{:.2f}'.format(score) for score in scores]))
```
We can keep the same test and it should work.

Goto15:
I reverted the requested code, but might have accidentally committed the change twice. My bad.

leouieda:
No problem, we've all done this at some point 🙂 

Thanks for contributing! It's much appreciated 🎉🎉🎉

PR #163, opened 2018-12-20
New median_distance func for near neighbors 

### LINKED ISSUE #127: Convenience function for near neighbor distances

Some applications would probably benefit from knowing the average distance between nearest data points (or the min, max, median, etc). For example, for determining the grid spacing or the `mindist` argument for `Spline`. 

This can be done with a KDTree:

```
from scipy.spatial import cKDTree
points = np.transpose([coordinates])
# Get the second closest (the closest are points themselves)
tree = cKDTree(points, k=[2])
distances, labels = tree.query(points)
distances.mean()
```

There is room for automation here to handle:

* The transposition of the coordinates and ensuring that there are only 2 1D arrays.
* Switching to pykdtree if it's installed.
* Projecting the coordinates.

A function like the following would be useful to have:

```
def neighbor_distances(coordinates, projection=None):
    ...
    return distances

spacing = neighbor_distances(coordinates).mean()
```

It could go in the `verde/coordinates.py` module.

leouieda:
The new `median_distance` function calculates the median near neighbor distance between each point in the given dataset. It automates the process of creating a kdtree and querying it on the same input points. It will take a median of the *k* nearest neighbors for each point and handle projections and reshaping of the output. 

This type of function is useful to get a sense for how scattered are the data points and what would be a reasonable grid spacing or block mean spacing.

Fixes #127 

**Reminders**


leouieda:
Actually, I'm gonna go ahead and merge this because I will need this function for some new gridders. But I'll wait a while before using it in the docs (#164).

PR #165, opened 2018-12-20
Utility function to create kdtrees

leouieda:
We support using pykdtree as a replacement for most kdtree operations
but the feature set is the not identical to the scipy implementation.
We had no way of dealing with this before and had some repeated
`try...except` code for importing the modules.

A new `kdtree` function in `verde/utils.py` centralizes the imports,
allows using the scipy kdtree even if pykdtree is installed,  and
handles transposing the coordinate arrays before creating the tree.

**Reminders**

### DIFF
--- a/verde/coordinates.py
+++ b/verde/coordinates.py
@@ -4,12 +4,8 @@ Functions for generating and manipulating coordinates.
 import numpy as np
 from sklearn.utils import check_random_state
 
-try:
-    from pykdtree.kdtree import KDTree
-except ImportError:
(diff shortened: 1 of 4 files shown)


djhoese:
Just curious, what features in the scipy kdtree are not supported in the pykdtree? Are there bugs in pykdtree that are making these things not work? Or did scipy add new features in the last year or two?

leouieda:
@djhoese no bugs, just the other class methods besides `query`: https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.cKDTree.html

I don't need them right now but I will need `query_ball_point` soon.

PR #181, opened 2019-04-17
Add enhanced support for geographic coordinates

### LINKED ISSUE #171: Improve support for geographic coordinates

Some functions from `verde` are intended to work with both projected and geographic coordinates.
Nevertheless, working with the latter imposes some requirements.
For example, one must ensure continuity around the globe: setting a grid between longitudes 170° and -170° or between 350° and 10° is not possible at the time because numerically the `east` coordinate is lower than the `west` one.
One way to fix this is to add a `latlon=True` argument to those functions in order to make clear that we are passing geographic coordinates.

I'll try to make a complete list of the functions that needs to be enhanced:
- `inside()`
- `grid_coordinates()`
- `scatter_coordinates()`
- `profile_coordinates()`
- `get_region()`
- `pad_region()`

santisoler:
Add `latlon_continuity()` function that moves a set of longitudinal coordinates either
to the `[0, 360)` or `[-180, 180)` degrees intervals, depending which one is more suited for
that particular set of coordinates.
It also allow to modify an extra set of coordinates based on the decision made on the
first set.
Modify the `inside()` function to enhance how the inside points are selected, even if
their longitudinal coordinates are not defined on the same degree interval. 

Fixes #171 

**Reminders**

### DIFF
--- a/verde/coordinates.py
+++ b/verde/coordinates.py
@@ -662,7 +662,7 @@ def inside(coordinates, region, latlon=False):
     """
     check_region(region, latlon=latlon)
     if latlon:
-        region, coordinates = latlon_continuity(region, coordinates=coordinates)
+        region, coordinates = longitude_continuity(region, coordinates=coordinates)
     w, e, s, n = region
     easting, northing = coordinates[:2]
     # Allocate temporary arrays to minimize memory allocation overhead
@@ -779,7 +779,7 @@ def block_split(coordinates, spacing=None, adjust="spacing", region=None, shape=
     return block_coords, labels
 
 
-def latlon_continuity(region, coordinates=None):
+def longitude_continuity(region, coordinates=None):
     """
     Modify geographic region boundaries to ensure continuity around the globe.
 
@@ -806,16 +806,16 @@ def latlon_continuity(region, coordinates=None):
     Examples
     --------
 
-    >>> from verde import latlon_continuity
+    >>> from verde import longitude_continuity
     >>> # Modify region with west > east
     >>> w, e, s, n = 350, 10, -10, 10
-    >>> print(latlon_continuity([w, e, s, n]))
+    >>> print(longitude_continuity([w, e, s, n]))
     [-10  10 -10  10]
     >>> # Modify region and extra coordinates
     >>> from verde import grid_coordinates
     >>> region = [-70, -60, -40, -30]
     >>> coordinates = grid_coordinates([270, 320, -50, -20], spacing=5)
-    >>> region, [longitude, latitude] = latlon_continuity(region, coordinates)
+    >>> region, [longitude, latitude] = longitude_continuity(region, coordinates)
     >>> print(region)
     [290 300 -40 -30]
     >>> print(longitude.min(), longitude.max())
@@ -823,7 +823,7 @@ def latlon_continuity(region, coordinates=None):
     >>> # Another example
     >>> region = [-20, 20, -20, 20]
     >>> coordinates = grid_coordinates([0, 350, -90, 90], spacing=10)
-    >>> region, [longitude, latitude] = latlon_continuity(region, coordinates)
+    >>> region, [longitude, latitude] = longitude_continuity(region, coordinates)
     >>> print(region)
     [-20  20 -20  20]
     >>> print(longitude.min(), longitude.max())
--- a/verde/coordinates.py
+++ b/verde/coordinates.py
@@ -753,11 +753,11 @@ def _latlon_continuity(west, east, longitude_coords):
 def _check_geographic_coordinates(coordinates):
     "Check if geographic coordinates are within accepted degrees intervals"
     longitude, latitude = coordinates[:2]
-    if (longitude > 360).all() or (longitude < -180).all():
+    if (longitude > 360).any() or (longitude < -180).any():
         raise ValueError(
-            "Invalid longitude coordinates. They should be < 360 and > -180 degrees"
+            "Invalid longitude coordinates. They should be < 360 and > -180 degrees."
         )
-    if (latitude > 90).all() or (latitude < -90).all():
+    if (latitude > 90).any() or (latitude < -90).any():
         raise ValueError(
-            "Invalid latitude coordinates. They should be < 90 and > -9 degrees"
+            "Invalid latitude coordinates. They should be < 90 and > -90 degrees."
         )
--- a/verde/coordinates.py
+++ b/verde/coordinates.py
@@ -8,7 +8,7 @@ from .base.utils import n_1d_arrays
 from .utils import kdtree
 
 
-def check_region(region, geographic=False):
+def check_region(region):
     """
     Check that the given region dimensions are valid.
 
@@ -20,8 +20,6 @@ def check_region(region, geographic=False):
     region : list = [W, E, S, N]
         The boundaries of a given region in Cartesian or geographic
         coordinates.
-    geographic : bool (optional)
-        If True the `region` will be assumed to be geographic coordinates in degrees.
 
     Raises
     ------
@@ -32,21 +30,12 @@ def check_region(region, geographic=False):
     if len(region) != 4:
         raise ValueError("Invalid region '{}'. Only 4 values allowed.".format(region))
     w, e, s, n = region
-    if geographic:
-        _check_geographic_coordinates([np.array([w, e]), np.array([s, n])])
-        if abs(e - w) > 360:
-            raise ValueError(
-                "Invalid region '{}' (W, E, S, N). ".format(region)
-                + "East and West boundaries must not be separated by an angle greater"
-                + "than 360 degrees."
-            )
-    else:
-        if w > e:
-            raise ValueError(
-                "Invalid region '{}' (W, E, S, N).Must have W =< E.".format(region)
-                + "If working with geographic coordinates, don't forget to add the "
-                + "geographic=True argument."
-            )
+    if w > e:
+        raise ValueError(
+            "Invalid region '{}' (W, E, S, N). Must have W =< E. ".format(region)
+            + "If working with geographic coordinates, don't forget to match geographic"
+            + " region with coordinates using `longitude_continuity` function."
+        )
     if s > n:
         raise ValueError(
             "Invalid region '{}' (W, E, S, N). Must have S =< N.".format(region)
@@ -603,7 +592,7 @@ def profile_coordinates(point1, point2, size, extra_coords=None):
     return tuple(coordinates), distances
 
 
-def inside(coordinates, region, geographic=False):
+def inside(coordinates, region):
     """
     Determine which points fall inside a given region.
 
@@ -618,9 +607,6 @@ def inside(coordinates, region, geographic=False):
     region : list = [W, E, S, N]
         The boundaries of a given region in Cartesian or geographic
         coordinates.
-    geographic : bool (optional)
-        If True both `region` and `coordinates` will be assumed to be geographic
-        coordinates in degrees.
 
     Returns
     -------
@@ -651,18 +637,17 @@ def inside(coordinates, region, geographic=False):
      [False  True  True]
      [False False False]]
     >>> # Geographic coordinates are also supported
+    >>> from verde import longitude_continuity
     >>> east, north = grid_coordinates([0, 350, -20, 20], spacing=10)
     >>> region = [-10, 10, -10, 10]
-    >>> are_inside = inside([east, north], region, geographic=True)
+    >>> are_inside = inside(*longitude_continuity([east, north], region))
     >>> print(east[are_inside])
     [  0.  10. 350.   0.  10. 350.   0.  10. 350.]
     >>> print(north[are_inside])
     [-10. -10. -10.   0.   0.   0.  10.  10.  10.]
 
     """
-    check_region(region, geographic=geographic)
-    if geographic:
-        region, coordinates = longitude_continuity(region, coordinates=coordinates)
+    check_region(region)
     w, e, s, n = region
     easting, northing = coordinates[:2]
     # Allocate temporary arrays to minimize memory allocation overhead
@@ -779,7 +764,7 @@ def block_split(coordinates, spacing=None, adjust="spacing", region=None, shape=
     return block_coords, labels
 
 
-def longitude_continuity(region, coordinates=None):
+def longitude_continuity(coordinates, region):
     """
     Modify geographic region boundaries to ensure continuity around the globe.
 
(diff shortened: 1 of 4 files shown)


leouieda:
@santisoler thanks for taking a shot at this! A few things to think about:

1. Does `latlon_continuity` need to be public? Doesn't seem like it will be very useful outside of `inside` or similar functions. It might be better to keep it private until we have a clear use case for it. 
2. This sort of thing is tricky and has loads of corner cases. I'd suggest making some more extensive tests (apart from doctests), particularly for corner cases (like regions around the poles or possibly overlapping on the other side of the globe like `-200/200`)

As a general rule, I find it useful to write the example for the feature first. That way I quickly see the usability problems and if I can't think of an example, then it probably isn't needed. It's always easier to make a private function public than the other way around.

santisoler:
Thanks @leouieda for the review.

>     1. Does `latlon_continuity` need to be public? Doesn't seem like it will be very useful outside of `inside` or similar functions. It might be better to keep it private until we have a clear use case for it.

I think you're right, it should be private, at least for now.
 
>     2. This sort of thing is tricky and has loads of corner cases. I'd suggest making some more extensive tests (apart from doctests), particularly for corner cases (like regions around the poles or possibly overlapping on the other side of the globe like `-200/200`)

Thanks for pointing this out. Writing test functions is tricky. First you think how the function **should work**, then spend time writing it and optimizing it. And then you have to switch your brain to **what could go wrong with it**. It's not easy work and we often need another pair of eyes.
I'll add more tests functions for such cases.

> As a general rule, I find it useful to write the example for the feature first. That way I quickly see the usability problems and if I can't think of an example, then it probably isn't needed. It's always easier to make a private function public than the other way around.

That's a nice way to start to write a new function. I'll try to implement it in the future.

santisoler:
@leouieda I've rewritten the `_lonlat_continuity` function. Initially I was thinking to be a general function that could be used by other ones besides `inside`, but so far it is the only function which makes use of it.
So I decided to simplify it. If other functions would need similar capabilities, we should leave that for the future.
Moreover, I rewrote some tests and add new ones, for example with overlapping coordinates and the cases on the poles.

I thought this enhancement would be straight forward, but each time I dive into it I find rough points.
Can you review it now?

santisoler:
@leouieda I have improved this codes based on we talked yesterday. I feel like it needs more testing, but it's looking better.
I have one issue with setting a region to `[350, -12, ..., ...]` and `[-10, -12, ..., ...]`. The first one raises an error due to the fact that `abs(-12 - 350) > 360`, while the second one does not gives errors, but the inside points are all `False` because `-10 > -12`.

I suspect that our constrains are the right ones, but we are not applying them in the right order.
I'll try to continue tackling down this.

verde/coordinates.py
```diff
@@ -702,3 +729,41 @@ def block_split(coordinates, spacing, adjust="spacing", region=None):
...
+        longitude = ((longitude + 180) % 360) - 180
+    region = [w, e, s, n]
+    coordinates = [longitude, latitude]
+    return coordinates, region
+
+
+def _check_geographic_coordinates(coordinates):
+    "Check if geographic coordinates are within accepted degrees intervals"
+    longitude, latitude = coordinates[:2]
+    if (longitude > 360).any() or (longitude < -180).any():
```
leouieda: Does `.any()` work with `pandas.Series`? If not, the maybe calling `np.any` might be better.

leouieda:
Sounds good, Santi. Let me know if you need feedback.

santisoler:
@leouieda I've made `latlon_continuity` public in order to make it available for other libraries like Harmonica.
Because it's now public, I've had to made region and coordinates checks inside it.
But I'm struggling how the `check_...` functions are written. They are quite ugly (not very easy to read) and, for example, the `inside()` function is calling the `check_region` twice: once is explicit while the other one is inside the `latlon_continuity`.
Maybe you could give me a fresh look on how to change them.

verde/coordinates.py
```diff
@@ -8,7 +8,7 @@
 from .utils import kdtree
 
 
-def check_region(region):
+def check_region(region, latlon=False):
```
leouieda: Would `geographic=False` be a better name? We already say "geographic" in the docs and comments to mean latitude and longitude.

verde/coordinates.py
```diff
@@ -702,3 +728,99 @@ def block_split(coordinates, spacing, adjust="spacing", region=None):
...
+            longitude = longitude % 360
+        else:
+            longitude = ((longitude + 180) % 360) - 180
+        coordinates = np.array(coordinates)
+        coordinates[0] = longitude
+        return region, coordinates
+    return region
+
+
+def _check_geographic_coordinates(coordinates):
```
leouieda: I don't think this function looks bad. Seems reasonable to me and does what it should.

verde/coordinates.py
santisoler: I like `geographic` more than the `latlon`. It improves readability and will be in agreement with the docs.

santisoler:
> 1. Since `latlon_continuity` is public, I think we don't need to call it inside of `inside`. This operation might be costly (if longitude coordinates need to be changed) and should be done at the start of a script, not inside each function that uses geographic data.

I agree. Maybe we could add a warning on `inside` docstring instructing to use `latlon_continuity` before using `inside` in case the user is working on geographic coordinates.

> 2. If that's the case, then the input and output of `latlon_continuity` should match `inside` so that we can do `inside(*latlon_continuity(coordinates, region))` or `coordinates, region = latlon_continuity(coordinates, region)` at the start of a script.
> 
>  3. The main functions of `latlon_continuity` are: 1) Make sure the longitude coordinates are continuous 2) adjust the region to match the coordinates. Discontinuous longitudes will cause problems when splitting the data, etc. It's better to change the region to match the coordinates than the other way around (the region is only 4 numbers).
> 
> 4. Given these goals, this function should always take `coordinates` as input. In the future, we could add an option to pass in `None` as coordinates if the need arises.

I disagree with these points. I think we don't have a precise goal for this function and this is causing some issues.
My first goal is to prepare the `region` and `coordinates` so `inside` works as expected even on geographic coordinates. Under that scenario, the `west` and `east` coordinates of `region` must be `west < east`. This obliged me to decide the best degree interval based on the `region` and then modify the `longitude` array to match it.

For example, if you have a global `longitude` array and a region around 0:
```python
longitude = np.linspace(0, 350, 36)
west, east = -10, 10
region = [west, east, south, north]
coordinates = [longitude, latitude]
```

If we modify the `region` based on the `longitude` we get:
```python
new_coordinates, new_region = latlon_continuity(coordinates, region)
print(np.all(new_coordinates == coordinates))
>> True
print(new_region[:2])
>> 350, 10
```
So, `west` is numerically greater than `east`. This will create problems on the `inside` function and it won't work as expected.
That's why I decided to adjust the `region` and then the `longitude` to match it.

> 5. Probably better to rename the function in this case. Maybe `longitude_conform` or `longitude_continuity` since the latitude is left unchanged.

That looks better than `latlon_...`. We should change it!

> This would probably solve a lot of the design issues you brought up. As I've mentioned before, when you find this type of problem it probably means that some of our initial design assumptions where wrong.
> 
> I like this new take a lot better and this would make this PR a lot more useful to other projects and applications.

You're right about how this function could be helpful for other projects.
Maybe we could start naming some requirements considering that this function should prepare the coordinates for `inside` and then see what could be improved for other applications.

verde/coordinates.py
santisoler: :+1:

santisoler:
@leouieda I've made some of the proposed changes.
I like how the code is looking now, although I should add a few more tests.
But I think the code is now cleaner and the usage is simpler.
The `longitude_continuity` accepts `coordinates=None` as argument, although I haven't added it to the docs.
Just wanted to know what do you think about letting this option, i.e.
```python
longitude_continuity(coordinates=None, region=region)
```

Please, take a look at it and let me know if I should change anything else or if I'm missing something.

verde/coordinates.py
```diff
@@ -702,3 +728,99 @@ def block_split(coordinates, spacing, adjust="spacing", region=None):
...
+    check_region([w, e, s, n], latlon=True)
+    # Check if region is defined all around the globe
+    all_globe = np.allclose(abs(e - w), 360)
+    # Move coordinates to [0, 360)
+    interval_360 = True
+    w = w % 360
+    e = e % 360
+    # Move west=0 and east=360 if region longitudes goes all around the globe
+    if all_globe:
+        w, e = 0, 360
```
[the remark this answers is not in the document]
santisoler: I've added this `all_globe` check in case you pass `w, e = 0, 360` as region boundaries.
If we omit the `all_globe`, the algorithm will return `w, e = 0, 0`, which is problematic.
The *slightly off coordinates** you mention will be within an error of 1e-5 (according to `numpy.allclose`).
One simple hack would be to lower down this error, but I don't know if it's ideal.

What do you think?

verde/coordinates.py
```diff
@@ -32,7 +32,9 @@ def check_region(region):
     w, e, s, n = region
     if w > e:
         raise ValueError(
-            "Invalid region '{}' (W, E, S, N). Must have W =< E.".format(region)
+            "Invalid region '{}' (W, E, S, N). Must have W =< E. ".format(region)
+            + "If working with geographic coordinates, don't forget to match geographic"
+            + " region with coordinates using `longitude_continuity` function."
```
leouieda:
```suggestion
            + " region with coordinates using 'verde.longitude_continuity'."
```

verde/coordinates.py
```diff
@@ -634,6 +636,15 @@ def inside(coordinates, region):
     [[False  True  True]
      [False  True  True]
      [False False False]]
+    >>> # Geographic coordinates are also supported
```
leouieda:
```suggestion
    Geographic coordinates are also supported using :func:`verde.longitude_continuity`:
    
```

verde/coordinates.py
```diff
@@ -751,3 +762,123 @@ def block_split(coordinates, spacing=None, adjust="spacing", region=None, shape=
     tree = kdtree(block_coords)
     labels = tree.query(np.transpose(n_1d_arrays(coordinates, 2)))[1]
     return block_coords, labels
+
+
+def longitude_continuity(coordinates, region):
+    """
+    Modify geographic region boundaries to ensure continuity around the globe.
```
leouieda:
```suggestion
    Modify coordinates and region boundaries to ensure longitude continuity.
```

verde/coordinates.py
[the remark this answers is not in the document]
leouieda: My thinking for this is "do we need to do this check?". If this is not necessary, then we could skip the check and avoid the issues with precision.

leouieda:
> Just wanted to know what do you think about letting this option, i.e.

Yep, the functionality looks good and the function does what we want. The only thing I'm entirely sure about is the function returning 1 or 2 things depending on input. I think this might come back to haunt us later. I'm leaning towards always returning `coordinates, region` even if `coordinates is None`. What do you think?

I think we talked about this but I can't remember: do we need `coordinates=None` right now or is this something that can be added later? If not, then I suggest we require coordinates and add this later if the need arises.

Do you have a sample dataset in mind that we could use for this? I can look for data around longitude 0 somewhere if not.

santisoler:
> Yep, the functionality looks good and the function does what we want. The only thing I'm entirely sure about is the function returning 1 or 2 things depending on input. I think this might come back to haunt us later. I'm leaning towards always returning `coordinates, region` even if `coordinates is None`. What do you think?

I have the same doubt regarding the 1 or 2 outputs. I think we should return `coordinates, region` on every case to avoid problems in the future.
We could add an example showing how to handle the `coordinate=None` case, something like:
```python
_, new_region = vd.longitude_continuity(None, region)
```

> I think we talked about this but I can't remember: do we need `coordinates=None` right now or is this something that can be added later? If not, then I suggest we require coordinates and add this later if the need arises.

Yep! The tesseroid implementation on Harmonica is a perfect example where you don't mind about the coordinates, you just want to adjust the boundaries of the tesseroid (remember that the computation point coordinates and tesseroid boundaries are *merged* inside the kernels, where the angles are always arguments of trigonometric functions, so we don't mind about phases).
 
> Do you have a sample dataset in mind that we could use for this? I can look for data around longitude 0 somewhere if not.

Any dataset could be useful. For example, ETOPO1 is on the `[-180, 180)` interval. We could make two examples from them:
1. Use a region on the `[0, 360)` interval, e.g. `[200, 240]`.
2. Then use a region around the 0 with the modified coordinates on the previous point.

Although I think finding a dataset defined on `[0, 360)` and try to use a region on `[-180, 180)` would be better.

verde/coordinates.py
```diff
@@ -751,3 +764,123 @@ def block_split(coordinates, spacing=None, adjust="spacing", region=None, shape=
     tree = kdtree(block_coords)
     labels = tree.query(np.transpose(n_1d_arrays(coordinates, 2)))[1]
     return block_coords, labels
+
+
+def longitude_continuity(coordinates, region):
+    """
+    Modify coordinates and region boundaries to ensure longitude continuity.
+
+    Longitudinal boundaries of the region are moved to the `[0, 360)` or `[-180, 180)`
```
leouieda:
```suggestion
    Longitudinal boundaries of the region are moved to the ``[0, 360)`` or ``[-180, 180)``
```

verde/coordinates.py
```diff
@@ -751,3 +764,123 @@ def block_split(coordinates, spacing=None, adjust="spacing", region=None, shape=
...
+    degrees interval depending which one is better suited for that specific region.
+
+    Parameters
+    ----------
+    coordinates : list or array
+        Set of geographic coordinates that will be moved to the same degrees
+        interval as the one of the modified region.
+    region : list or array
+        List or array containing the boundary coordinates `w, `e`, `s`, `n` of the
```
leouieda: Missing extra `

leouieda:
Hi @santisoler, left a few comments here with minor things. What do we need to get this merged? The API looks good to me and the functionally seems to all be there. The test coverage is missing a fit and there are some CI errors. We could try to move this forward soon to make a new Verde release.

santisoler:
@leouieda Now `coordinates.py` meet the 100% coverage and `pylint` doesn't not raise any errors.
The Azure Windows build is complaining because `test_spline_warns_underdetermined` fails. I'm almost sure this PR doesn't break anything related with spline, but I'm might be wrong.

Let me know if something else should be done here, otherwise I think this PR it's ready to be merged!

leouieda:
> I'm almost sure this PR doesn't break anything related with spline, but I'm might be wrong.

That is unrelated and a bit of a bad test now that I think about it. I don't know why it's complaining so much actually. Probably an update of numba that issues a new warning. I'll see if I can fix this in another PR before merging this. If not, then I'll merge as is.

Thanks for implementing this!

Issue #161, opened 2018-12-17
Tutorial for grid generation and coordinates


leouieda:
**Description of the desired feature**

We need a tutorial about generating regular grids and all the details surrounding how the coordinates are determined (spacing, adjustments, registration, etc). This would one of the very first tutorials and would help explain some of the concepts in the `grid_coordinates` docstring in more detail.

jessepisel:
I'm thinking of taking a crack at doing some documentation in June. I think this would be a fun place to start with a tutorial that incorporates #164 in as well. @leouieda What do you think? Also, do you have any particular verde dataset in mind?

leouieda:
@jessepisel that would be great! All help is much appreciated :)  No dataset in mind. Actually, this one probably doesn't need a dataset since it won't be gridding anything. It's more about the details of how we handle grid coordinates generation (spacing vs shape, adjusting spacing vs region, pixel registration, etc). If needed, I think `verde.datasets.CheckerBoard` should do nicely.

Not sure if #164 fits here, though. It might be better to have another page with the "Basics of Interpolation" or something like that.

jessepisel:
That sounds good to me. I will go through and make a tutorial for grid generation and leave #164 for later on this month. I'll put in a PR when I have something this upcoming week.

PR #192, opened 2019-06-21
Tutorial about how grid coordinates are generated

jessepisel:
I created a tutorial for grid generation that covers region and spacing adjustment and pixel registration.  Let me know if this is what you had in mind for this tutorial and I can change as needed. Addresses #161 

**Reminders**

- [X] Run `make format` and `make check` to make sure the code follows the style guide.
- [na ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [na] Add new public functions/methods/classes to `doc/api/index.rst` and `verde/__init__.py`.
- [na] Write detailed docstrings for all functions/classes/methods.
- [X] If adding new functionality, add an example to the docstring, gallery, and/or tutorials.


leouieda:
Thanks, Jesse! I'll have a look at this soon. Meanwhile, @santisoler could you have a look at the tutorial since you're not as familiar with the implementation? You might spot something that I won't.

tutorials/grid_coordinates.py
```diff
@@ -0,0 +1,208 @@
...
+
+########################################################################################
+# How Grid Coordinates works
+# --------------------------
+# When we choose to adjust the region the function takes the original region, and
+# divides the south-north and west-east extents of the region by the spacing to
+# get the number of grid points in both directions. This value is then
+# rounded and converted to an integer. The function then adds one to the integer
+# to get the number of nodes rather than segments.The function then multiplies
+# the number of grid points by the chosen spacing to create the new region.
```
santisoler: I don't know if explaining what exactly is going on under the hood is entirely necessary on this kind of tutorials.
Maybe would be nicer to show the problem that raises from the initial choice of region and spacing.
Something like:

```
We cannot define a set of regularly spaced points by 500 units on this region:
the range of the west-east and south-north boundaries are not multiple of 500 units.
So we must choose to change either:
- the boundaries of the region in order to fit the spacing, or
- the spacing in order to fit the region boundaries.
```

Then we can introduce the two approaches: adjust region or adjust spacing.

tutorials/grid_coordinates.py
```diff
@@ -0,0 +1,208 @@
...
+print(regioneast.shape, regionnorth.shape)
+
+########################################################################################
+# With the spacing set at 500 units and a 3 by 4 grid of regular dimensions,
+# :func:`verde.grid_coordinates` then calculates the spatial location of each
+# grid point and adjusts the region so that the maximum northing and maximum
+# easting values are divisible by the spacing. In this example, the easting has
+# 3 segments (4 nodes) that are each 500 units long, meaning the easting spans
+# from 0 to 1500. The northing has 2 segments (3 nodes) that are each 500 units
+# long, meaning the northing spans from 0 to 1000. Both dimensions are divisble
```
santisoler:
```suggestion
# long, meaning the northing spans from 0 to 1000. Both dimensions are divisible
```

tutorials/grid_coordinates.py
```diff
@@ -0,0 +1,208 @@
...
+plt.show()
+
+######################################################################################
+# Pixel Registration
+# ------------------
+# Pixel registration locates the grid points in the middle of the grid segments
+# rather than in the corner of each grid.
+# First, let's take the same region and grid, and set the `adjust` parameter to
+# ``region`` so that the function will adjust the region, and set
+# ``pixel_register`` parameter to `true`. Without piexel registration our grid
```
santisoler:
```suggestion
# ``pixel_register`` parameter to `True`. Without pixel registration our grid
```

tutorials/grid_coordinates.py
```diff
@@ -0,0 +1,208 @@
...
+====================
+
+Grid coordinates in Verde are used to create points on a regularly spaced grid
+that are  then used in the spline method to interpolate between sample data
+points. The grid can be specified either by the number of points in each
+dimension (the *shape*) or by the grid node spacing.
+
+Creating regular grids in Verde is done using the :func:`verde.grid-coordinates`
+function. The function uses :func:`numpy.linspace` and
+:func:`numpy.meshgrid` to interpolate between points in both the west-east and
```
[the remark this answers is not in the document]
jessepisel: I agree with you on this

tutorials/grid_coordinates.py
jessepisel: I like how you reframed this section. I think it's easier to understand the difference between adjusting spacing vs boundaries.

jessepisel:
@santisoler Thank you so much for the review! I accepted all the typos and missing markdown styling changes you made. 
For the rest of your comments:
> I would like to introduce the tutorial with a non problematic region and spacing (e.g. `w, e, s, n = 0, 1000, 0, 1000` and `spacing = 100`) just to show what `grid_coordinates` creates. And then show the one you proposed. What do you think?

I like this idea, I think that if it starts with an easy to understand and non problematic spacing it would give a better understanding of what `grid_coordinates` is doing. 
 
> Regarding the plots, have you tried adding some transparency to the points? Doing so may help visualizing if the two grids overlap. Besides, how do you feel adding a grid to the plots, maybe with the same spacing as one of the computed grids? I think it would help to visualize the slight differences between both grids.

I think transparency on the points would help a ton, and grids would be really useful for visualizing the grid. Great suggestion!

> Regarding the pixel registration, I would keep the region or spacing adjust aside. I would choose a non-problematic combination of region and spacing and show how the differences between grids created with `pixel_register = True` and `pixel_register = False`. I'm just thinking to focus on what pixel registration does independently of the previous adjustment.

This would be great to show the difference between pixel registration and normal gridding. It should help people on that front as well. 

I will update this later this week and commit a new version after I have made your suggested changes!

santisoler:
I'm glad you found my suggestions useful!
Let me know if you need help or more reviewing. :+1:

jessepisel:
@santisoler I updated the tutorial with an initial non-problematic grid, and added some grid lines to the other plots. Will you take a look at it and see if it makes sense the way it laid out?

tutorials/grid_coordinates.py
```diff
@@ -0,0 +1,324 @@
...
+
+Gridding coordinates
+====================
+
+Grid coordinates in Verde are used to create points on a regularly spaced grid
+that are then used in the spline method to interpolate between sample data
+points. The grid can be specified either by the number of points in each
+dimension (the *shape*) or by the grid node spacing.
+
+Creating regular grids in Verde is done using the :func:`verde.grid_coordinates`
```
santisoler: I wouldn't mention the adjusting of region or spacing until next section. So I would simplify this paragraph as follows:
```
Creating regular grids in Verde is done using the :func:`verde.grid_coordinates`
function. It creates a set of regulary spaced points in both the west-east and
south-north directions, i.e. a two-dimensional spatial grid.
```

tutorials/grid_coordinates.py
```diff
@@ -0,0 +1,324 @@
...
+function. The function uses :func:`numpy.linspace` and
+:func:`numpy.meshgrid` to create a set of regulary spaced points in both the
+west-east and south-north directions to create a two-dimensional spatial grid.
+To this end, the function will adjust either the region bounds or the spacing
+between grid nodes.
+
+First let's create a region that is 1000 units west-east and 1000 units south-
```
santisoler: The same for this paragraph:
```
First let's create a region that is 1000 units west-east and 1000 units south-
north, and we will set an initial spacing to 100 units.
```

tutorials/grid_coordinates.py
```diff
@@ -0,0 +1,324 @@
...
+
+plt.ylim(-50, 1050)
+plt.xlim(-50, 1050)
+plt.xlabel("Easting")
+plt.ylabel("Northing")
+plt.legend(loc="upper center", bbox_to_anchor=(0.5, 1.15))
+plt.show()
+
+########################################################################################
+# How Grid Coordinates works
```
santisoler: I would simplify these three sections ("How Grid Coordinates works", "Create the grid with adjust region" and "Create the grid with adjust spacing") into two: maybe "Adjusting region boundaries when creating the grid" and "Adjusting spacing when creating the grid".

So maybe it could look like this:
```python
# Adjusting region boundaries when creating the grid
# --------------------------------------------------
# Now let's create a region that is 1555 units west-east, and 1250 units south-north
# with a spacing of 500 units. Because the range of the west-east and south-north
# boundaries are not multiples of 500, we must choose to change either:
#   - the boundaries of the region in order to fit the spacing, or
#   - the spacing in order to fit the region boundaries.
#
# We could tell :func:`verde.grid_coordinates` to adjust the region boundaries by
# passing `adjust="region"`.

region_east, region_north = vd.grid_coordinates(
    region=region, spacing=spacing, adjust="region"
)
print(region_east.shape, region_north.shape)

...
```

```python
# Adjusting spacing when creating the grid
# ----------------------------------------
# Now let's adjust the spacing of the grid points by passing `adjust="spacing"` to
# :func:`verde.grid_coordinates`.

spacing_east, spacing_north = vd.grid_coordinates(
    region=region, spacing=500, adjust="spacing"
)
print(spacing_east.shape, spacing_north.shape)

...
```

What do you think?

tutorials/grid_coordinates.py
jessepisel: I have simplified this section and moved adjustments to the next sections

tutorials/grid_coordinates.py
jessepisel: Also simplified

tutorials/grid_coordinates.py
jessepisel: I think that three sections with your proposed headers make a lot of sense, and have a solid flow to them. I have changed them in the tutorial

jessepisel:
Thanks @santisoler I appreciate the review. I like all of your changes for the organization, I think it makes sense working through it. The headers for each section fit nicely as well. I also added the section on `extra_coords`, I totally missed that when I was going through the doc string. I added a short section with `extra_coords` for both a height and a time. Thanks for the review again, take a look at what I changed and let me know if it works!

tutorials/grid_coordinates.py
leouieda: I agree with @santisoler. We probably don't need the first paragraph. One more distinction, `grid_coordinates` makes the *coordinates* for a grid, not the grid iteself.

tutorials/grid_coordinates.py
```diff
@@ -0,0 +1,346 @@
+"""
+.. _grid_coordinates:
+
+Gridding coordinates
```
leouieda:
```suggestion
Grid coordinates
```

tutorials/grid_coordinates.py
leouieda: The line break at the `-` makes results in a space when rendered to html

leouieda: I would move this paragraph to after the `import` block so that the text and variables/functions are closer together.

tutorials/grid_coordinates.py
```diff
@@ -0,0 +1,346 @@
...
+import verde as vd
+from matplotlib.patches import Rectangle
+
+# create the bounds for the region and set the spacing for the grid points
+spacing = 100
+west, east, south, north = 0, 1000, 0, 1000
+region = (west, east, south, north)
+
+# create the grid
+square_region_east, square_region_north = vd.grid_coordinates(
```
leouieda: I really like that you're using nice and descriptive variable names! In this case, though, it might distract the reader as to what these actually mean. I would suggest renaming them to `easting` and `northing` instead. What do you think?

tutorials/grid_coordinates.py
```diff
@@ -0,0 +1,346 @@
...
+First let's create a region that is 1000 units west-east and 1000 units south-
+north, and we will set an initial spacing to 100 units.
+"""
+
+import numpy as np
+import matplotlib.pyplot as plt
+import verde as vd
+from matplotlib.patches import Rectangle
+
+# create the bounds for the region and set the spacing for the grid points
```
leouieda: Part of the reason for moving the paragraph above closer is that it will make the comments obsolete (so it's easier to maintain updated).

tutorials/grid_coordinates.py
```diff
@@ -0,0 +1,346 @@
...
+west, east, south, north = 0, 1000, 0, 1000
+region = (west, east, south, north)
+
+# create the grid
+square_region_east, square_region_north = vd.grid_coordinates(
+    region=region, spacing=spacing
+)
+
+########################################################################################
+# We can check the dimensions of the grid to confirm that it is 11x11
```
leouieda:
```suggestion
# The region can divided into exactly 10x10 segments, resulting in
# 11x11 grid points.
# We can check the dimensions of the grid to confirm that it is 11x11
```

tutorials/grid_coordinates.py
```diff
@@ -0,0 +1,346 @@
...
+
+print(square_region_east.shape, square_region_north.shape)
+
+########################################################################################
+# Now let's plot the grid to visualize the location of the grid points
```
leouieda:
```suggestion
# Now let's visualize the location of the grid points
```

tutorials/grid_coordinates.py
```diff
@@ -0,0 +1,346 @@
...
+    color="black",
+    s=100,
+)
+
+plt.ylim(-50, 1050)
+plt.xlim(-50, 1050)
+plt.xlabel("Easting")
+plt.ylabel("Northing")
+plt.legend(loc="upper center", bbox_to_anchor=(0.5, 1.15))
+plt.show()
```
leouieda: These plots are great! Would it make sense to have this code in a function since it's repeated below? It could something like:

```python
def plot_grid(ax, coordinates, linestyles="dotted", region=None, pad=50, **kwargs):
    "Plot the grid coordinates as dots and lines."
    if region is not None:
        west, east, south, north = region
        ax.add_patch(
            plt.Rectangle((west, south), east, north, fill=None, label="Region Bounds")
        )
    data_region = vd.get_region(coordinates)
    ax.vlines(
        coordinates[0][0],
        ymin=data_region[2],
        ymax=data_region[3],
        linestyles=linestyles,
        zorder=0,
    )
    ax.hlines(
        coordinates[1][:, 1],
        xmin=data_region[0],
        xmax=data_region[1],
        linestyles=linestyles,
        zorder=0,
    )
    ax.scatter(*coordinates, **kwargs)
    if pad:
        padded = vd.pad_region(region, pad=pad)
        plt.xlim(padded[:2])
        plt.ylim(padded[2:])
```

This way the plotting can be simplified to this later on:

```python

plt.figure(figsize=(6, 6))
plot_grid(
    ax=plt.subplot(111), 
    coordinates=(square_region_east, square_region_north), 
    region=region,
    label="Square Region Grid Nodes",
    marker=".",
    color="black",
    s=100,
)
plt.xlabel("Easting")
plt.ylabel("Northing")
plt.legend(loc="upper center", bbox_to_anchor=(0.5, 1.15))
plt.show()
```

tutorials/grid_coordinates.py
```diff
@@ -0,0 +1,346 @@
...
+
+########################################################################################
+# Adjusting region boundaries when creating the grid
+# --------------------------------------------------
+# Now let's create a region that is 1555 units west-east, and 1250 units south-north
```
leouieda: It might be better to keep the region the same as the previous example and modify the spacing slightly so that it doesn't divide exactly anymore. This way, between the last example and this one only 1 thing changes (keeping the cognitive burden low).

tutorials/grid_coordinates.py
santisoler: Add a missing *be* before *divided*.

leouieda:
@jessepisel thanks for all this great work! There are changes now in the master branch that fix all those failing tests. Let me know when you're ready and we can merge in the master branch here.

jessepisel:
@leouieda sounds great! I have an updated version of the tutorial that addresses your comments. I am having some errors when I build the documentation, should I commit and then merge master, or merge master and then commit the new changes?

leouieda:
@jessepisel feel free to commit your changes and then merge master. In general, feel free to commit as much as you want. When we merge the pull request, all commits get squashed down into one and we rewrite the commit message based on the PR description. So the history will always be clean even if you have many broken commits in the PR.

jessepisel:
@leouieda Perfect. I have added the latest commit and merged from master. Good to know the commit messages are rewritten when the PR is merged, I wasn't sure what the process was for message history. Thank you again for all the help!

tutorials/grid_coordinates.py
```diff
@@ -0,0 +1,289 @@
...
+====================
+
+Grid coordinates in Verde are used to create points on a regularly spaced grid
+that are then used in the spline method to interpolate between sample data
+points. The grid can be specified either by the number of points in each
+dimension (the *shape*) or by the grid node spacing.
+
+Creating regular grids in Verde is done using the :func:`verde.grid_coordinates`
+function. It creates a set of regularly spaced points in both the west-east and
+south-north directions, i.e. a two-dimensional spatial grid.
```
leouieda: @jessepisel how about this slight rephrasing of the two paragraphs above (they would be merged into a single one):

Creating the coordinates for regular grids in Verde is done using the :func:\`verde.grid_coordinates\` function. It creates a set of regularly spaced points in both the west-east and south-north directions, i.e. a two-dimensional spatial grid. These points are then used by the Verde gridders to interpolate between data points. As such, all \`.grid\` methods (like :meth:\`verde.Spline.grid\`) take as input the configuration parameters for :func:\`verde.grid_coordinates\`. The grid can be specified either by the number of points in each dimension (the *shape*) or by the grid node spacing.

tutorials/grid_coordinates.py
```diff
@@ -0,0 +1,289 @@
...
+
+spacing = 100
+west, east, south, north = 0, 1000, 0, 1000
+region = (west, east, south, north)
+
+# create the grid coordinates
+easting, northing = vd.grid_coordinates(region=region, spacing=spacing)
+
+########################################################################################
+# We can check the dimensions of the grid coordinates to confirm that it is 11x11
```
leouieda: It would be good to explain why it should be 11x11.

tutorials/grid_coordinates.py
```diff
@@ -0,0 +1,289 @@
...
+
+print(easting.shape, northing.shape)
+
+########################################################################################
+# Let's define two functions to visualize the region bounds and grid points
+
+
+def plot_region(ax, region=None):
+    "Plot the region as a solid line."
+    if region is not None:
```
leouieda: Since this function only plots the region, there is no need for the `region=None` default.

tutorials/grid_coordinates.py
```diff
@@ -0,0 +1,289 @@
...
+        zorder=0,
+    )
+    ax.scatter(*coordinates, **kwargs)
+    if pad:
+        padded = vd.pad_region(region, pad=pad)
+        plt.xlim(padded[:2])
+        plt.ylim(padded[2:])
+
+
+# Visualize our region and grid coordinates using our functions
```
leouieda: Maybe add a `############...` to make this render as rst instead of a comment?

tutorials/grid_coordinates.py
```diff
@@ -0,0 +1,289 @@
...
+# --------------------------------------------------------------
+#
+# Now let's change our spacing to 300 units. Because the range of the west-east and
+# south-north boundaries are not multiples of 300, we must choose to change either:
+#
+# - the boundaries of the region in order to fit the spacing, or
+# - the spacing in order to fit the region boundaries.
+#
+# We could tell :func:`verde.grid_coordinates` to adjust the region boundaries by
+# passing `adjust="region"`.
```
leouieda:
```suggestion
# passing ``adjust="region"``.
```

tutorials/grid_coordinates.py
```diff
@@ -0,0 +1,289 @@
...
+# easting values are divisible by the spacing. In this example, both the  easting and
+# northing have 3 segments (4 nodes) that are each 300 units long, meaning the easting
+# and northing span from 0 to 900. Both dimensions are divisible
+# by 300.
+
+print(region_easting)
+print(region_northing)
+
+######################################################################################
+# By default if adjust is not assigned to ``region`` or ``spacing``,
```
leouieda:
```suggestion
# By default, if ``adjust`` is not assigned to ``"region"`` or ``"spacing"``,
```

tutorials/grid_coordinates.py
```diff
@@ -0,0 +1,289 @@
...
+# in a similar manner as when it adjusts the region. However, it doesn't readjust
+# the region so that it is divisble by the spacing before creating the grid.
+# This means the grid will have the same number of grid points no matter if
+# the adjust parameter is set to ``region`` or ``spacing``.
+
+########################################################################################
+# Adjusting spacing when creating the grid
+# ----------------------------------------
+#
+# Now let's adjust the spacing of the grid points by passing `adjust="spacing"`
```
leouieda:
```suggestion
# Now let's adjust the spacing of the grid points by passing ``adjust="spacing"``
```

tutorials/grid_coordinates.py
```diff
@@ -0,0 +1,289 @@
...
+plt.ylabel("Northing")
+plt.legend(loc="upper center", bbox_to_anchor=(0.5, 1.18))
+plt.show()
+
+######################################################################################
+# Pixel Registration
+# ------------------
+#
+# Pixel registration locates the grid points in the middle of the grid segments
+# rather than in the corner of each grid.
```
leouieda:
```suggestion
# rather than in the corner of each grid node.
```

tutorials/grid_coordinates.py
```diff
@@ -0,0 +1,289 @@
...
+#
+# First, let's take our 1000x1000 region and use the 100 unit spacing from the first
+# example and set the ``pixel_register`` parameter to `True`. Without pixel
```
leouieda:
```suggestion
# example and set the ``pixel_register`` parameter to ``True``. Without pixel
```

tutorials/grid_coordinates.py
```diff
@@ -0,0 +1,289 @@
...
+
+######################################################################################
+# And we can check the coordinates for the grid points with pixel registration.
+
+print(pixel_easting)
+print(pixel_northing)
+
+######################################################################################
+# If we set ``pixel_register`` to ``False`` the function will return the grid
+# with one more grid node in both west-east and south-north directions.
```
leouieda:
```suggestion
# coordinates of the nodes instead of pixel centers, resulting in an extra point in each direction.
```

tutorials/grid_coordinates.py
jessepisel: I like this change, it integrates the gridders into the tutorial and lets the user know why they are using `grid_coordinates`

tutorials/grid_coordinates.py
jessepisel: added a brief explanation for the shape (segments vs nodes)

tutorials/grid_coordinates.py
jessepisel: removed the default

tutorials/grid_coordinates.py
jessepisel: I like it as rst better than an inline comment. Looks much better!

leouieda:
@jessepisel great work! This is ready to merge on my part. 

@santisoler do you have anything else to add? If not, we can merge this in and release v1.2.0.

leouieda:
Merging this in. Thanks again @jessepisel!

PR #220, opened 2020-01-16
Use pooch.os_cache and pkg_resources in datasets

leouieda:
With Pooch 0.7.0, the recommended way of loading the registry file is
with `pkg_resources` (see fatiando/pooch#120). It's also better to use
the default cache location so users can more easily clean up unused
files. Because this is system specific, add the `verde.datasets.locate`
function to return the cache folder location.

**Reminders**


leouieda:
Fixed an issue in the CI (was setting the environment variable wrong). Will submit the same fix to Harmonica as well.

PR #221, opened 2020-01-16
Explicitly set numba config for each jit function

leouieda:
Using `parallel=True` on the Green's function was raising a warning
because there is nothing to parallelize. Set arguments individually to
avoid this. Add missing `numba.prange` to Spline Jacobian calculation to
actually run it in parallel.

**Reminders**


leouieda:
I think #222 might fix some of these failures so waiting on that.

leouieda:
For some reason, the Dask related tests and docs hang in CI when using numba. I can't reproduce locally so it's hard to pin point. Will try removing the `parallel` arg from numba to see if that's the problem.

leouieda:
Alright, using the numpy engine with the dask tests seems to work. Strange because the tutorials work.

leouieda:
Merging this in since things seem to be working.

PR #229, opened 2020-02-26
Bug fix in grid_to_table for wrong order of Dataset coords

leouieda:
We were relying on the order of the `coords` attribute of the data set to `meshgrid` the coordinates. This is wrong because xarray takes the coordinate order from `dims` instead, which is what we should also do. Took this chance to refactor the code and included a test that would have caught the bug.

**Reminders**:


leouieda:
Thanks, @jessepisel! Merging this in now.

Issue #173, opened 2019-03-15
Interpolate data on global geographic coordinates


santisoler:
**Description of the desired feature**

The current interpolators in Verde are not able to interpolate data on global geographic coordinates.
This is a not just a software problem, but rather a mathematical one.
[Wessel and Becker (2008)](http://doi.org/10.1111/j.1365-246X.2008.03829.x) present a method to interpolate global data based on Green's functions and splines.
Would be nice to have this implemented on Verde.

**Are you willing to help implement and maintain this feature?** Yes/No

leouieda:
It would indeed be very good to have this in there! I imagine you could pretty much copy the `Spline` class and change the Green's functions. The controlling parameters are also a bit different so it might not be worth the effort of subclassing. This will be a nice test to see if we should redesign `Spline` to be more subclass friendly.

PR #231, opened 2020-02-26
Fix bug with profile distances being unprojected

leouieda:
`BaseGridder.profile` has the option to apply a projection to the
coordinates before predicting so we can pass geographic coordinates to
cartesian gridders. In these cases, the distance along the profile is
calculated by `profile_coordinates` with the unprojected coordinates (in
the geographic case it would be degrees). Returning unprojected
distances is not very useful and instead we should have been returning
projected distances. They need to be calculated from the projected
coordinates (out of `profile_coordinates`) because the points are evenly
spaced in unprojected coordinates.

**Reminders**:


leouieda:
@santisoler @jessepisel I could use another set of eyes on this one. Using the example of providing geographic points to `profile` and a projection:

1. Generate profile points in degrees before projecting (so they are evenly spaced in degrees).
2. Project the coordinates and predict.
3. Re-calculate distances between the projected coordinates so we don't return "distances" in degrees.

Does this seem reasonable? I found this bug while extracting a profile for modelling in 2D Cartesian coordinates and getting distances in degrees (which are kind of useless). Then again, maybe I should have just projected the coordinates of the points before passing them to `profile`. What do you think?

leouieda:
Actually, it might make more sense to project the points before creating the profile. The way we were generating the profile coordinates is probably wrong if given geographic coordinates.

santisoler:
IMHO, I think we should avoid projecting inside the method and ask already projected coordinates.
This would make the workflow more transparent and it will increase users' code just by one line (projecting the coordinates, the projection has to be defined on both cases).
Worth remembering that projections may introduce errors if the projected region is too big, so I think we should avoid introducing such source of error inside the `profile` method.

If we would like to create profiles on geographic coordinates maybe we could write a different method for that, which don't involve projecting coordinates but computing equidistant points located on a sphere. But that's for another PR of course.

santisoler:
It worth noting that this workflow would work if the data region is not too big. Otherwise we will be introducing errors due to projection. A solution for this would be to create profiles directly on geographic (or spherical) coordinates, where equidistant points can be defined as point located at constant arc-length distances. But this should be tackled on a different PR.

ahartikainen:
What library are you using to calculate distances? Would it be a good idea to calculate distances with `pyproj` where it calculates the geographic distance?

leouieda:
@santisoler yeah I agree that this is not the best workflow. The main problem is that the gridder isn't aware of coordinate systems. They are all Cartesian at the moment as well so the problem with large areas and profiles is present for all our gridding. 

@ahartikainen we're not really using any libraries for the reasons outlines above. Right now, we just assume that everything is Cartesian. We could use pyproj for great circle distances but that might involve having `profile` be aware of Cartesian vs geographic (which it currently isn't).

It seems more and more to be that we can't use the same abstraction for `profile` and `grid` (pass in the projection and don't ask questions). We might have to add an option to inform profile that we're talking about geographic coordinates.

The ideal case for geographic profiles would be:

1. Make the profile evenly spaced in actual physical coordinates (not degrees).
2. Return distances as great circle distances instead of "degrees".

When extracting profiles, we're usually plotting value x distance so it's more useful to have things work this way most times.

leouieda:
A few thoughts after discussion with @santisoler:

* The Verde gridders are currently all Cartesian by nature.
* Even if they take a `projection` argument, the actual prediction is still happening in Cartesian coordinates.
* So it makes little sense to calculate distances on the sphere and then flatten it out for interpolation.
* In case we have spherical gridders in the future (#173), then it would make sense for those gridders to calculate distances on the sphere by default. That would require changes to how we generate the profile coordinates and it will likely have to be different than for the Cartesian case. We can deal with that when the need arises. 

I think the most sensible approach right now is:

* Project the profile points before passing them to `profile_coordinates`. That way, the profile coordinates are generated in Cartesian coordinates and so are the distances.
* Do the inverse projection of the profile coordinates before sending them back to the user. That way the profile location can be plotted on a map.

This would make the profile point generation be Cartesian (as it should for the current gridders) and solve the problem of returning "distances" as angles.

@ahartikainen @santisoler what do you think?

leouieda:
I implemented projecting the input points and then projecting back the Cartesian coordinates. This way, the distances returned are Cartesian (projected) and evenly spaced in the projection domain. I also added a section in the projections tutorial about profiles:

2020-03-17-111057_3200x1800_scrot

My main question right now is that this breaks compatibility slightly since the returned points and distances are going to be different from before. But I consider the old behaviour to be a bug so it might not be too bad.

If we consider this to break the API compatibility, I'd be in favour of holding it until Verde 2.0. If not, then we can merge as include it in 1.4.0, which I want to release once #237 is done.

What do you think?

cc @santisoler @ahartikainen @jessepisel

santisoler:
This is looking great @leouieda!

I agree to merge this. Although it somehow breaks backward compatibility, I agree the previous implementation was bugged, so I think we can prioritize accurate behavior over backward compatibility. Maybe we could post a comment about this on Slack after we merge this so most users can become aware of this. In fact we could ask their opinion before merging.

leouieda:
> Although it somehow breaks backward compatibility, I agree the previous implementation was bugged, so I think we can prioritize accurate behavior over backward compatibility. 

Yep, and also change the values of coordinates but not the number/types of arguments returned or any API so we should be fine. This is also not a very well publicized method and likely not very widely used.

Another option is to include a warning about the change which we can remove in 2.0.

leouieda:
Added more details to the docstring and a warning (docs only) about the change.

leouieda:
Merging. Thank you all for the suggestions and reviews :)

PR #236, opened 2020-03-11
Add function for rolling windows on irregular data

leouieda:
The `rolling_window` function returns the indices of points falling on
rolling/moving windows. It works for irregularly sampled data and is a
nice complement to `xarray.DataArray.rolling`  (which only works on
grids). Uses a KDTree to select points but is not compatible with
pykdtree (it's missing the method we need).

**Reminders**:

### DIFF
--- a/verde/coordinates.py
+++ b/verde/coordinates.py
@@ -799,8 +799,7 @@ def rolling_split(
     ----------
     coordinates : tuple of arrays
         Arrays with the coordinates of each data point. Should be in the
-        following order: (easting, northing, vertical, ...). Only easting and
-        northing will be used, all subsequent coordinates will be ignored.
+        following order: (easting, northing, vertical, ...).
     size : float
         The size of the windows. Units should match the units of *coordinates*.
     spacing : float, tuple = (s_north, s_east), or None
@@ -946,6 +945,7 @@ def rolling_split(
     centers = grid_coordinates(
         window_region, spacing=spacing, shape=shape, adjust=adjust
     )
+    step = centers[0][0]
     # pykdtree doesn't support query_ball_point yet and we need that
     tree = kdtree(coordinates, use_pykdtree=False)
     # Coordinates must be transposed because the kd-tree wants them as columns
--- a/verde/coordinates.py
+++ b/verde/coordinates.py
@@ -948,6 +948,9 @@ def rolling_split(
     )
     # pykdtree doesn't support query_ball_point yet and we need that
     tree = kdtree(coordinates, use_pykdtree=False)
+    # Coordinates must be transposed because the kd-tree wants them as columns
+    # of a matrix
+    # Use p=inf (infinity norm) to get square windows instead of circular ones
     indices1d = tree.query_ball_point(
         np.transpose(n_1d_arrays(centers, 2)), r=size / 2, p=np.inf
     )
--- a/verde/coordinates.py
+++ b/verde/coordinates.py
@@ -735,6 +735,7 @@ def block_split(coordinates, spacing=None, adjust="spacing", region=None, shape=
     See also
     --------
     BlockReduce : Apply a reduction operation to the data in blocks (windows).
+    rolling_split : Split the given points on a rolling (moving) window.
 
     Examples
     --------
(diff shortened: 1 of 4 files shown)


leouieda:
They idea for this is to do rolling operations on irregular data. In gravmag, the main usage is for Euler deconvolution (which is what I plan on using this for). I thought of implementing something like `DataArray.rolling` which is much more sophisticated, but there is no way to make it general enough for our purposes. So for now it just returns the indices of points for each window. 

Any thoughts?

verde/coordinates.py
```diff
@@ -770,15 +771,186 @@ def block_split(coordinates, spacing=None, adjust="spacing", region=None, shape=
...
+    ]
+    centers = grid_coordinates(
+        window_region, spacing=spacing, shape=shape, adjust=adjust
+    )
+    # pykdtree doesn't support query_ball_point yet and we need that
+    tree = kdtree(coordinates, use_pykdtree=False)
+    indices1d = tree.query_ball_point(
+        np.transpose(n_1d_arrays(centers, 2)), r=size / 2, p=np.inf
+    )
+    indices = [np.unravel_index(i, shape=shapes[0]) for i in indices1d]
```
santisoler: This line is throwing an error if one of the rolling windows is unpopulated.
For example, if I run the following code:
```python
import verde as vd

coordinates = vd.scatter_points(region=(0, 100, 0, 100), size=30, random_state=1234)
vd.rolling_split(coordinates, size=20, spacing=10)
```

I get the following:
```traceback
Traceback (most recent call last):
  File "rolling.py", line 13, in <module>
    vd.rolling_split(coordinates, size=size, spacing=10)
  File "/home/santi/git/verde/verde/coordinates.py", line 953, in rolling_split
    indices = [np.unravel_index(i, shape=shapes[0]) for i in indices1d]
  File "/home/santi/git/verde/verde/coordinates.py", line 953, in <listcomp>
    indices = [np.unravel_index(i, shape=shapes[0]) for i in indices1d]
  File "<__array_function__ internals>", line 6, in unravel_index
TypeError: indices must be integral: the provided empty sequence was inferred as float. Wrap it with 'np.array(indices, dtype=np.intp)'
```

I think converting `i` to a `np.array` with `dtype=int` should solve it:
```suggestion
    indices = [
        np.unravel_index(np.array(i, dtype=int), shape=shapes[0])
        for i in indices1d
    ]
```

verde/coordinates.py
```diff
@@ -770,15 +771,186 @@ def block_split(coordinates, spacing=None, adjust="spacing", region=None, shape=
...
+    window_region = [
+        dimension + (-1) ** (i % 2) * size / 2 for i, dimension in enumerate(region)
+    ]
+    centers = grid_coordinates(
+        window_region, spacing=spacing, shape=shape, adjust=adjust
+    )
+    # pykdtree doesn't support query_ball_point yet and we need that
+    tree = kdtree(coordinates, use_pykdtree=False)
+    indices1d = tree.query_ball_point(
+        np.transpose(n_1d_arrays(centers, 2)), r=size / 2, p=np.inf
```
santisoler: Would be nice to comment this line, explaining why are we transposing the window centers.
Also would be nice to explain in a comment what does it mean to use a `p=np.inf` Minkowski norm and how we need to use it instead of another norm.

leouieda: Saw this coming yesterday :wink: https://github.com/fatiando/verde/pull/238/files#diff-c495cd4f5ca0b2b47ab963a1deae043dR897 Will add it here as well.

verde/coordinates.py
leouieda: Right, that seems reasonable. Thanks!

verde/coordinates.py
```diff
@@ -770,15 +771,186 @@ def block_split(coordinates, spacing=None, adjust="spacing", region=None, shape=
...
+    Parameters
+    ----------
+    coordinates : tuple of arrays
+        Arrays with the coordinates of each data point. Should be in the
+        following order: (easting, northing, vertical, ...). Only easting and
+        northing will be used, all subsequent coordinates will be ignored.
+    spacing : float, tuple = (s_north, s_east), or None
+        The window size in the South-North and West-East directions,
+        respectively. A single value means that the size is equal in both
+        directions.
```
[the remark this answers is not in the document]
leouieda: Oops

leouieda:
@santisoler that's spot on, thanks! I'll add the warnings (exceptions would be a bit too strict since there is nothing particularly wrong about non-overlapping windows) and some tests to make sure they are raised. Thank you for the review :+1:

leouieda:
@santisoler implemented your suggestions and this is ready for another round of reviews.

The only thing I left out is checking if `size == spacing`. The output of `block_split` is actually a bit different so they 2 functions aren't entirely the interchangeable. `block_split` returns a label for each block while `rolling_split` returns the indices for each window. So even if `size==spacing`, `block_split` returns a single array that has be used in a `groupby` operation while `rolling_split` returns multiple arrays used for indexing.

santisoler:
You're welcome @leouieda! 

> I'll add the warnings (exceptions would be a bit too strict since there is nothing particularly wrong about non-overlapping windows) and some tests to make sure they are raised.

That's ok for me. I cannot think of a use case where we want to use non-overlapping windows (and thus leaving _unwindowed_ data), but you're right it's not a problem to have them implemented.

> The only thing I left out is checking if size == spacing. The output of block_split is actually a bit different so they 2 functions aren't entirely the interchangeable. block_split returns a label for each block while rolling_split returns the indices for each window. So even if size==spacing, block_split returns a single array that has be used in a groupby operation while rolling_split returns multiple arrays used for indexing.

You're right, they are not exactly making the same thing. I was misled by the similar names.

leouieda:
OK, a TODO for this:

- [ ] Rename to `rolling_window` to avoid confusion with `block_split` since they are fundamentally different
- [ ] See if it's possible to preserve the shape of the window center coordinates by reshaping the list of indices.

leouieda:
@santisoler done with the changes. What do you think? I was able to preserve the shape of the windows arrays.

leouieda:
Merging away :rocket:

PR #237, opened 2020-03-11
Add a convex hull masking function

### LINKED ISSUE #126: Create a convexhull masking function

A good way to mask grid points that are too far from data is to only show the ones that fall inside the data convex hull. This is what many scipy and matplotlib interpolations do. It would be great to have a `convexhull_mask` function that has a similar interface to `distance_mask` but masks points outside of the convex hull. This function should take the same arguments as `distance_mask` except `maxdist`. 

One way of implementing this would be with the `scipy.spatial.Delaunay` class:

```
tri = Delaunay(np.transpose(data_coordinates))
# Find which triangle each grid point is in. -1 indicates that it's not in any.
in_triangle = tri.find_simplex(np.transpose(coordinates))
mask = in_triangle > 0
```

leouieda:
Use scipy's Delaunay triangulation to determine points that are outside
the convex hull of the data points. Useful to mask out wildly
extrapolated points from grids. It's easier to use than `distance_mask`
because it doesn't require distance calculations (and the associated
projections). Use the new function in some examples and tutorials.

Fixes #126

**Reminders**:

### DIFF
--- a/verde/mask.py
+++ b/verde/mask.py
@@ -109,7 +109,7 @@ def distance_mask(
 
 
 def convexhull_mask(
-    data_coordinates, coordinates=None, grid=None,
+    data_coordinates, coordinates=None, grid=None, projection=None,
 ):
     """
     Mask grid points that are outside the convex hull of the given data points.
@@ -134,6 +134,13 @@ def convexhull_mask(
         the grid as northing and easting coordinates, respectively. The mask
         will be applied to *grid* using the :meth:`xarray.Dataset.where`
         method.
+    projection : callable or None
+        If not None, then should be a callable object ``projection(easting,
+        northing) -> (proj_easting, proj_northing)`` that takes in easting and
+        northing coordinate arrays and returns projected easting and northing
+        coordinate arrays. This function will be used to project the given
+        coordinates (or the ones extracted from the grid) before calculating
+        distances.
 
     Returns
     -------
@@ -176,13 +183,27 @@ def convexhull_mask(
 
     """
     coordinates, shape = _get_grid_coordinates(coordinates, grid)
-    n_coordinates = len(data_coordinates)
-    triangles = Delaunay(np.transpose(n_1d_arrays(data_coordinates, n_coordinates)))
+    n_coordinates = 2
+    # Make sure they are arrays so we can normalize
+    data_coordinates = n_1d_arrays(data_coordinates, n_coordinates)
+    coordinates = n_1d_arrays(coordinates, n_coordinates)
+    if projection is not None:
+        data_coordinates = projection(*data_coordinates)
+        coordinates = projection(*coordinates)
+    # Normalize the coordinates to avoid errors from qhull when values are very
+    # large (as occurs when projections are used).
+    means = [coord.mean() for coord in data_coordinates]
+    stds = [coord.std() for coord in data_coordinates]
+    data_coordinates = tuple(
+        (coord - mean) / std for coord, mean, std in zip(data_coordinates, means, stds)
+    )
+    coordinates = tuple(
+        (coord - mean) / std for coord, mean, std in zip(coordinates, means, stds)
+    )
+    triangles = Delaunay(np.transpose(data_coordinates))
     # Find the triangle that contains each grid point.
     # -1 indicates that it's not in any triangle.
-    in_triangle = triangles.find_simplex(
-        np.transpose(n_1d_arrays(coordinates, n_coordinates))
-    )
+    in_triangle = triangles.find_simplex(np.transpose(coordinates))
     mask = (in_triangle != -1).reshape(shape)
     if grid is not None:
         return grid.where(mask)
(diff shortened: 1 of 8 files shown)


verde/mask.py
```diff
@@ -110,3 +106,100 @@ def distance_mask(
...
+     [nan nan  1.  1. nan nan]
+     [nan nan  1.  1. nan nan]
+     [nan nan  1.  1. nan nan]
+     [nan nan nan nan nan nan]
+     [nan nan nan nan nan nan]]
+
+    """
+    coordinates, shape = _get_grid_coordinates(coordinates, grid)
+    n_coordinates = len(data_coordinates)
+    triangles = Delaunay(np.transpose(n_1d_arrays(data_coordinates, n_coordinates)))
```
ahartikainen: Does delaunay still use qhull library which uses 32 floats (or something else weird) at some point of it's processing pipeline. This forces one to use "normalized" coordinates, otherwise points are truncated to a "grid"

leouieda: That's a good question. From the docs, I would say "yes": https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.Delaunay.html#scipy.spatial.Delaunay but they don't say anything about that. 

What do you mean by "normalized"? Scale them to [0, 1]? If you have any pointers that would be great :)

leouieda: I found this page but there is nothing on that: http://www.qhull.org/html/qh-impre.htm

leouieda: By the way, thanks for finding these things Ari. Always things I would never have thought to look for :+1:

ahartikainen: Normalized --> small enough

I once did stuff with 1000 m x 1000 m area (points in random position) with real coordinates (northern sweden, some coordinate system with large values) and had this problem. It took sometime to figure out why results were funny.

Good thing is that can be done only for the data going in the Delaunay and then use the original data for rest of the workflow.

Maybe some simple script with large offset from origo could tell if this is still a problem or not.

verde/mask.py
```diff
@@ -110,3 +106,100 @@ def distance_mask(
...
+def _get_grid_coordinates(coordinates, grid):
+    """
+    If coordinates is given, return it and their shape. Otherwise, get
+    coordinate arrays from the grid.
+    """
+    if coordinates is None and grid is None:
+        raise ValueError("Either coordinates or grid must be given.")
+    if coordinates is None:
+        dims = [grid[var].dims for var in grid.data_vars][0]
+        coordinates = np.meshgrid(grid.coords[dims[1]], grid.coords[dims[0]])
```
santisoler: By choosing the `northing` and `easting` based on the order of `dims` could be a source of errors if the dimensions are not in the expected order.

verde/mask.py
leouieda: Thanks for the pointers! I'll try to add a test for that to see if it's something we need to worry about.

verde/mask.py
leouieda: I thought of this after #229 but we're getting dims from the `DataArray` so ii should be fine. The only thing that might cause problems is if the dims are different for the different vars. I could add a check for that and error if that's the case. 

This is why adding these things back to xarray is hard. I don't think I could generalize this to arbitrary `Dataset`.

santisoler: The problem may appear if the user is using a different order on the `dims` of the `DataArray`. I've seen that after applying some `xarray` functions, sometimes the order of `dims` are changed (sorry, don't remember exactly which ones, but found [this issue](https://github.com/pydata/xarray/issues/2811) that may be related).
I don't have a final solution for this, just wanted to raise the discussion.

leouieda: Right, I hadn't thought of that. It might be worth mentioning in the docs that we're assuming this order of dimensions then.

verde/mask.py
ahartikainen: https://gist.github.com/ahartikainen/bf2c0eb2ed493040ff490e0c97435ade

Not so sure if this can also fail for convex hull, probably it can

leouieda: Ah, that's severe actually. It probably will since the process I'm using is checking if points fall on any simplex. I'm add this and see if the gallery example changes much.

leouieda: I added the normalization code and support for projecting the coordinates prior to the masking so we can get really large coordinates. This is the gallery example without normalization:

qhull

And this is after normalization:

qhull-norm

So no difference in this case. I'll leave the normalization code in there since it doesn't seem to hurt the results and might prevent headaches in the future.

leouieda:
TODO:

- [ ] Test case with projection
- [ ] Mention in the docs that we assume that xarray dims are northing, easting

leouieda:
Should be done with this. @santisoler @ahartikainen if you take another look that would be great. As always, no rush on this :)

leouieda:
Merged! Thank you for the reviews @santisoler and @ahartikainen

PR #240, opened 2020-03-12
Generalize BaseGridder docstrings

### LINKED ISSUE #239: Generalize BaseGridder docstrings

Class `verde.BaseGridder` is used to create a broad type of data interpolators.
When subclassing it, we just need to define only certain methods for each specific type of interpolation, while many other methods are used directly from the originals ones defined on `verde.BaseClass`. Most of them are compatible with any coordinate system (Cartesian, spherical or geodetic), except for `profile` (which involves computing distances) (see #231 for more dicussion about this). Nevertheless, the docstrings of some of these methods aren't so general, and some of them specifically asks variables defined in Cartesian coordinates. This may create inconsistent docstrings for subclasses that are intended to work with a specific coordinate system (see fatiando/harmonica#157).

A possible solution is to make the `verde.BaseGridder` docstrings more general, so the methods are not asking for a specific coordinate system. Also, we could add the `dims` argument of `grid` method as a variable class, so when we subclass it we can override it.

santisoler:
Make docstrings of `BaseGridder` more general, without specifying
a coordinate system. Add `dims` as a class variable that will be used by
default on `grid`, `scatter` and `profile` methods.

Fixes #239

**Reminders**:

### DIFF
--- a/verde/base/base_classes.py
+++ b/verde/base/base_classes.py
@@ -422,9 +422,12 @@ class BaseGridder(BaseEstimator):
         size : int
             The number of points to generate.
         dims : list or None
-            The names of the northing and easting data dimensions,
-            respectively, in the output dataframe. Defaults to ``['northing',
-            'easting']``. **NOTE: This is an exception to the "easting" then
+            The names of the northing/latitude and easting/longitude data
+            dimensions, respectively, in the output grid. Defaults are set
+            though ``dims`` class variable. Dims must be defined in the
+            following order: ``["northing", "easting"]`` or
+            ``["latitude", "longitude"]``.
+            **NOTE: This is an exception to the "easting" then
             "northing" pattern but is required for compatibility with xarray.**
         data_names : list of None
             The name(s) of the data variables in the output dataframe. Defaults
--- a/verde/base/base_classes.py
+++ b/verde/base/base_classes.py
@@ -82,7 +82,7 @@ class BaseGridder(BaseEstimator):
     -32.2, -32.2, -32.2, -32.2, -32.2, -32.2, -32.2, -32.2, -32.2, -32.2
 
     """
-
+# The default dimension names for generated outputs (pd.DataFrame, xr.Dataset, etc)
     dims = ("northing", "easting")
 
     def predict(self, coordinates):
--- a/verde/base/base_classes.py
+++ b/verde/base/base_classes.py
@@ -82,7 +82,9 @@ class BaseGridder(BaseEstimator):
     -32.2, -32.2, -32.2, -32.2, -32.2, -32.2, -32.2, -32.2, -32.2, -32.2
 
     """
-# The default dimension names for generated outputs (pd.DataFrame, xr.Dataset, etc)
+
+    # The default dimension names for generated outputs
+    # (pd.DataFrame, xr.Dataset, etc)
     dims = ("northing", "easting")
 
     def predict(self, coordinates):
@@ -149,9 +151,9 @@ class BaseGridder(BaseEstimator):
         ----------
         coordinates : tuple of arrays
             Arrays with the coordinates of each data point. Should be in the
-            following order: (easting, northing, vertical, ...). 
-            For the specific definition of coordinate systems and what these 
-            names mean, see the class docstring. 
+            following order: (easting, northing, vertical, ...).
+            For the specific definition of coordinate systems and what these
+            names mean, see the class docstring.
         data : array or tuple of arrays
             The data values of each data point. If the data has more than one
             component, *data* must be a tuple of arrays (one for each
--- a/verde/base/base_classes.py
+++ b/verde/base/base_classes.py
@@ -262,10 +262,10 @@ class BaseGridder(BaseEstimator):
             The grid spacing in the South-North and West-East directions,
             respectively.
         dims : list or None
-            The names of the northing/latitude and easting/longitude data
-            dimensions, respectively, in the output grid. Default is determined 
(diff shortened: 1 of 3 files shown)


santisoler:
@leouieda Would you like to take a look at this? Is this what you had in mind?

verde/base/base_classes.py
```diff
@@ -83,6 +83,8 @@ class BaseGridder(BaseEstimator):
 
     """
```
leouieda:
```suggestion
# The default dimension names for generated outputs (pd.DataFrame, xr.Dataset, etc)
```

verde/base/base_classes.py
```diff
@@ -147,7 +149,9 @@ def filter(self, coordinates, data, weights=None):
         ----------
         coordinates : tuple of arrays
             Arrays with the coordinates of each data point. Should be in the
-            following order: (easting, northing, vertical, ...).
+            following order: (easting, northing, vertical, ...)
+            or (longitude, latitude, height, ...)
+            or (longitude, latitude, radius, ...).
```
leouieda: This might get a bit confusing. How about this:

```suggestion
            following order: (easting, northing, vertical, ...). 
            For the specific definition of coordinate systems and what these 
            names mean, see the class docstring. 
```
Then the class can define that the system is spherical and these mean longitude, latitude, radius, for example.

verde/base/base_classes.py
```diff
@@ -246,18 +252,20 @@ def grid(
         Parameters
         ----------
         region : list = [W, E, S, N]
-            The boundaries of a given region in Cartesian or geographic
-            coordinates.
+            The boundaries of a given region.
```
leouieda:
```suggestion
            The west, east, south, and north boundaries of a given region.
```

verde/base/base_classes.py
```diff
@@ -246,18 +252,20 @@ def grid(
...
         dims : list or None
-            The names of the northing and easting data dimensions,
-            respectively, in the output grid. Defaults to ``['northing',
-            'easting']``. **NOTE: This is an exception to the "easting" then
+            The names of the northing/latitude and easting/longitude data
+            dimensions, respectively, in the output grid. Defaults are set
+            though ``dims`` class variable. Dims must be defined in the
+            following order: ``["northing", "easting"]`` or
+            ``["latitude", "longitude"]``.
+            **NOTE: This is an exception to the "easting" then
```
leouieda:
```suggestion
            The names of the northing/latitude and easting/longitude data
            dimensions, respectively, in the output grid. Default is determined 
            from the ``dims`` attribute of the class. Must be defined in the
            following order: northing dimension, easting dimension.
            **NOTE: This is an exception to the "easting" then
```

I don't think we need to add "longitude" and "latitude" everywhere. That is bound to get confusing and hard to keep consistent. I expect users know that easting => longitude and northing => latitude (which is why I like those names as generic alternatives).

verde/base/base_classes.py
```diff
@@ -267,7 +275,8 @@ def grid(
             3D vector data.
         projection : callable or None
             If not None, then should be a callable object
-            ``projection(easting, northing) -> (proj_easting, proj_northing)``
+            ``projection(longitude, latitude) ->
+            (proj_easting, proj_northing)``
```
leouieda: Not necessarily. The projection could be reversed for a spherical gridder that we want to make a Cartesian grid. That's why it's hard to be too specific in these classes.

verde/base/base_classes.py
```diff
@@ -451,22 +466,23 @@ def profile(
         return pd.DataFrame(dict(columns), columns=[c[0] for c in columns])
 
 
-def get_dims(dims):
+def get_dims(gridder, dims):
```
leouieda: Might be better off as a private method in this case.

verde/tests/test_base.py
```diff
@@ -18,8 +18,11 @@
 
 def test_get_dims():
     "Tests that get_dims returns the expected results"
-    assert get_dims(dims=None) == ("northing", "easting")
-    assert get_dims(dims=("john", "paul")) == ("john", "paul")
+    gridder = BaseGridder()
+    assert get_dims(gridder, dims=None) == ("northing", "easting")
+    assert get_dims(gridder, dims=("john", "paul")) == ("john", "paul")
+    gridder.dims = ("latitude", "longitude")
+    assert get_dims(gridder, dims=None) == ("latitude", "longitude")
```
leouieda: This test is a bit useless since it does the same thing as the doctest. Could probably get rid of it.

verde/base/base_classes.py
santisoler: Agree

verde/base/base_classes.py
santisoler: Done!

verde/tests/test_base.py
santisoler: Now `_get_dims` is a small private method, so I don't see the point of having an example for it. Wouldn't be better to remove the example and keep the tests?

leouieda: In the end, it doesn't matter that much. They both test the same thing and it's just a matter of having it as a doctest or regular test. The advantage of the doctest is that it serves as internal documentation.

santisoler:
I'm quite liking to create examples that somehow test the behavior of the method/function. In fact it's a nice way to create quick tests when drafting a new function. Nevertheless, for this case I decided to remove the examples from `_get_dims` just because it's a little private method and also very easy to read.

santisoler:
@leouieda feel free to merge if CIs pass.

Issue #108, opened 2018-08-03
Collaboration with pyresample


djhoese:
I am one of the maintainers of the pyresample project (https://github.com/pytroll/pyresample) along with @mraspaud @adybbroe @pnuu. We do some very similar operations to what you have in verde, but with different interfaces and sometimes different purposes. I think our two projects could benefit from each other.

Pyresample/SatPy
------------------

Pyresample has been around for a long time and is used by many meteorological institutes/organizations via the satpy library (https://github.com/pytroll/satpy) and mpop before that. I also use satpy in my command line tools under the polar2grid and geo2grid project and joined the PyTroll/SatPy team after attending conferences and realizing that we were solving exactly the same problems.

Pyresample and satpy now use xarray and dask for all of their operations, but pyresample is in need of a new interface now that xarray accessors are more popular and the ease of attaching geolocation information to an xarray object (see https://github.com/pydata/xarray/issues/2288). Pyresample offers (or will very soon) three different resampling algorithms: nearest neighbor using pykdtree, elliptical weighted averaging for scan-based satellite instrument data, and bilinear interpolation using pykdtree for neighbor calculations. The old pyresample API also includes the ability to use your own weighting functions, but we haven't migrated that to work with the xarray/dask functionality.

The use cases that pyresample were originally used for and what people were used to doing involved having predefined 500m-2km grids in various projections. These grids can also be dependent on the visualization tool used to view the satellite data where only certain projections can be used or certain resolutions are too dense for satellite raster images. This is my opinion/experience, the other pyresample developers may see differently. Anyway, a lot of pyresample's interfaces assume you have a predefined grid ("AreaDefinition") that you want to resample your data to. It is coming up more and more that we need an easier interface for making these definitions from other existing definitions like lon/lat arrays or adjustments to existing AreaDefinitions (change resolution). While this is all possible, it isn't always the easiest to do or access or document.

Verde
------

I spoke with you @leouieda at SciPy 2018 and on gitter which lead to #92 where verde now uses pykdtree if it is available for better performance. You also mentioned in your lightning talk I believe that you'd like to switch to dask where possible but haven't done so yet. This is something pyresample's experience could help with since dask-ifying a kdtree is not easy, but I think we've done the best we can without rewriting the C code.

[...]

PR #246, opened 2020-03-26
Add function to project 2D gridded data

leouieda:
The `project_grid` function transforms a grid to the given projection. It re-samples the data using `ScipyGridder` (by default) and runs a blocked mean (optional) to avoid aliasing when the points aren't evenly distributed in the projected coordinates (like in polar projections). Applies a `convexhul_mask` to the grid always to avoid extrapolation to points that had no original data (this is done by scipy for linear and cubic interpolation already). 

The function only works for `xarray.DataArray`s (not `Dataset`) because I couldn't find an easy way to guarantee that a variables have the same dimensions. We can add that option later on if needed. It wouldn't break the API so we can use this as a minimum viable product for now.

Move the new function and `project_region` to the new module `verde/projections.py`.

**Reminders**:

- [x] Add new public functions/methods/classes to `doc/api/index.rst` and the base `__init__.py` file for the package.
- [x] Write detailed docstrings for all functions/classes/methods. It often helps to design better code if you write the docstrings first.
- [x] Add your full name, affiliation, and [ORCID](https://orcid.org) (optional) to the `AUTHORS.md` file (if you haven't already) in case you'd like to be listed as an author on the [Zenodo](https://zenodo.org/communities/fatiando) archive of the next release.

### DIFF
--- a/examples/project_grid.py
+++ b/examples/project_grid.py
@@ -1,0 +1,74 @@
+"""
+Projection of gridded data
+==========================
+
+Sometimes gridded data products need to be projected before they can be used.
+For example, you might want to project the topography of Antarctica from
+geographic latitude and longitude to a planar polar stereographic projection
+before doing your analysis. When projecting, the data points will likely not
+fall on a regular grid anymore and must be interpolated (re-sampled) onto a new
+grid.
+
+Function :func:`verde.project_grid` automates this process using the
+interpolation methods available in Verde. An input grid
+(:class:`xarray.DataArray`) is interpolated onto a new grid in the given
+`pyproj <https://jswhit.github.io/pyproj/>`__ projection. The function takes
+care of choosing a default grid spacing and region, running a blocked mean to
+avoid spatial aliasing (using :class:`~verde.BlockReduce`), and masking the
+points in the new grid that aren't constrained by the original data (using
+:func:`~verde.convexhull_mask`).
+
+In this example, we'll generate a synthetic geographic grid with a checkerboard
+pattern around the South pole. We'll project the grid to South Polar
+Stereographic, revealing the checkered pattern of the data.
+
+.. note::
+
+    The interpolation methods are limited to what is available in Verde and
+    there is only support for single 2D grids. For more sophisticated use
+    cases, you might want to try
+    `pyresample <https://github.com/pytroll/pyresample>`__ instead.
+
+"""
+import numpy as np
+import matplotlib.pyplot as plt
+import pyproj
+import verde as vd
+
+
+# We'll use synthetic data near to South pole to highlight the effects of the
+# projection. EPSG 3031 is a South Polar Stereographic projection.
+projection = pyproj.Proj("epsg:3031")
+
+# Create a synthetic geographic grid using a checkerboard pattern
+region = (0, 360, -90, -60)
+spacing = 0.25
+wavelength = 10 * 1e5  # The size of the cells in the checkerboard
+checkerboard = vd.datasets.CheckerBoard(
+    region=vd.project_region(region, projection), w_east=wavelength, w_north=wavelength
+)
+data = checkerboard.grid(
+    region=region,
+    spacing=spacing,
+    projection=projection,
+    data_names=["checkerboard"],
+    dims=("latitude", "longitude"),
+)
+
+# Do the projection setting the output grid spacing (in projected meters). Set
+# the coordinates names to x and y since they aren't really "northing" or
+# "easting".
+polar_data = vd.project_grid(
+    data.checkerboard, projection, spacing=0.5 * 1e5, dims=("y", "x")
+)
+
+# Plot the original and projected grids
+fig, (ax1, ax2) = plt.subplots(1, 2, figsize=(10, 6))
+data.checkerboard.plot(
+    ax=ax1, cbar_kwargs=dict(orientation="horizontal", aspect=50, pad=0.1)
+)
+ax1.set_title("Geographic Grid")
+polar_data.plot(ax=ax2, cbar_kwargs=dict(orientation="horizontal", aspect=50, pad=0.1))
+ax2.set_title("Polar Stereographic Grid")
+plt.tight_layout()
+plt.show()
--- a/examples/project_grid.py
+++ b/examples/project_grid.py
@@ -54,6 +54,8 @@ data = checkerboard.grid(
     data_names=["checkerboard"],
     dims=("latitude", "longitude"),
 )
+print("Geographic grid:")
+print(data)
 
 # Do the projection setting the output grid spacing (in projected meters). Set
 # the coordinates names to x and y since they aren't really "northing" or
@@ -61,6 +63,8 @@ data = checkerboard.grid(
 polar_data = vd.project_grid(
     data.checkerboard, projection, spacing=0.5 * 1e5, dims=("y", "x")
 )
+print("\nProjected grid:")
+print(polar_data)
 
 # Plot the original and projected grids
 fig, (ax1, ax2) = plt.subplots(1, 2, figsize=(10, 6))
--- a/examples/project_grid.py
+++ b/examples/project_grid.py
@@ -9,7 +9,7 @@ before doing your analysis. When projecting, the data points will likely not
 fall on a regular grid anymore and must be interpolated (re-sampled) onto a new
 grid.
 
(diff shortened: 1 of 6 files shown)


leouieda:
This is what the gallery example looks like:

```python
import numpy as np
import matplotlib.pyplot as plt
import pyproj
import verde as vd

# We'll use synthetic data near to South pole to highlight the effects of the
# projection. EPSG 3031 is a South Polar Stereographic projection.
projection = pyproj.Proj("epsg:3031")

# Create a synthetic geographic grid using a checkerboard pattern
region = (0, 360, -90, -60)
spacing = 0.25
wavelength = 10 * 1e5  # The size of the cells in the checkerboard
checkerboard = vd.datasets.CheckerBoard(
    region=vd.project_region(region, projection), w_east=wavelength, w_north=wavelength
)
data = checkerboard.grid(
    region=region,
    spacing=spacing,
    projection=projection,
    data_names=["checkerboard"],
    dims=("latitude", "longitude"),
)
print("Geographic grid:")
print(data)

# Do the projection setting the output grid spacing (in projected meters). Set
# the coordinates names to x and y since they aren't really "northing" or
# "easting".
polar_data = vd.project_grid(
    data.checkerboard, projection, spacing=0.5 * 1e5, dims=("y", "x")
)
print("\nProjected grid:")
print(polar_data)

# Plot the original and projected grids
fig, (ax1, ax2) = plt.subplots(1, 2, figsize=(10, 6))
data.checkerboard.plot(
    ax=ax1, cbar_kwargs=dict(orientation="horizontal", aspect=50, pad=0.1)
)
ax1.set_title("Geographic Grid")
polar_data.plot(ax=ax2, cbar_kwargs=dict(orientation="horizontal", aspect=50, pad=0.1))
ax2.set_title("Polar Stereographic Grid")
plt.tight_layout()
plt.show()
```

sphx_glr_project_grid_001

leouieda:
@djhoese I was hoping to get your input on this PR as it overlaps with pyresample quite a bit (reviving our discussion in #108). This is a lot simpler than what pyresample can do but it's often what I need and not worth the effort of setting up the `AreaDefinition` etc in pyresample. But I'd like to bring this to your attention as a possible way to collaborate :slightly_smiling_face: 

I'm particularly interested in the parallel kdtree and dask support in pyresample (but that's beyond this PR).

jessepisel:
@leouieda This looks like a great contribution. I will go ahead with a code review tomorrow and see what I can find.

leouieda:
Thanks @jessepisel, I really appreciate it :slightly_smiling_face:

examples/project_grid.py
```diff
@@ -0,0 +1,78 @@
...
+==========================
+
+Sometimes gridded data products need to be projected before they can be used.
+For example, you might want to project the topography of Antarctica from
+geographic latitude and longitude to a planar polar stereographic projection
+before doing your analysis. When projecting, the data points will likely not
+fall on a regular grid anymore and must be interpolated (re-sampled) onto a new
+grid.
+
+Function :func:`verde.project_grid` automates this process using the
```
jessepisel, line 12:
```suggestion
The :func:`verde.project_grid` function automates this process using the
```

examples/project_grid.py
```diff
@@ -0,0 +1,78 @@
...
+interpolation methods available in Verde. An input grid
+(:class:`xarray.DataArray`) is interpolated onto a new grid in the given
+`pyproj <https://jswhit.github.io/pyproj/>`__ projection. The function takes
+care of choosing a default grid spacing and region, running a blocked mean to
+avoid spatial aliasing (using :class:`~verde.BlockReduce`), and masking the
+points in the new grid that aren't constrained by the original data (using
+:func:`~verde.convexhull_mask`).
```
jessepisel, line 19: This is a great description of how it works and how the function deals with the original data. I dig it!

examples/project_grid.py
```diff
@@ -0,0 +1,78 @@
...
+pattern around the South pole. We'll project the grid to South Polar
+Stereographic, revealing the checkered pattern of the data.
+
+.. note::
+
+    The interpolation methods are limited to what is available in Verde and
+    there is only support for single 2D grids. For more sophisticated use
+    cases, you might want to try
+    `pyresample <https://github.com/pytroll/pyresample>`__ instead.
+
```
jessepisel, line 31: This is a useful note with link to pyresample

examples/project_grid.py
```diff
@@ -0,0 +1,78 @@
...
+"""
+import numpy as np
+import matplotlib.pyplot as plt
+import pyproj
+import verde as vd
+
+
+# We'll use synthetic data near to South pole to highlight the effects of the
```
jessepisel, line 39:
```suggestion
# We'll use synthetic data near the South pole to highlight the effects of the
```

examples/project_grid.py
```diff
@@ -0,0 +1,78 @@
...
+    region=region,
+    spacing=spacing,
+    projection=projection,
+    data_names=["checkerboard"],
+    dims=("latitude", "longitude"),
+)
+print("Geographic grid:")
+print(data)
+
+# Do the projection setting the output grid spacing (in projected meters). Set
```
jessepisel, line 60:
```suggestion
# Do the projection while setting the output grid spacing (in projected meters). Set
```

examples/project_grid.py
```diff
@@ -0,0 +1,78 @@
...
+# Plot the original and projected grids
+fig, (ax1, ax2) = plt.subplots(1, 2, figsize=(10, 6))
+data.checkerboard.plot(
+    ax=ax1, cbar_kwargs=dict(orientation="horizontal", aspect=50, pad=0.1)
+)
+ax1.set_title("Geographic Grid")
+polar_data.plot(ax=ax2, cbar_kwargs=dict(orientation="horizontal", aspect=50, pad=0.1))
+ax2.set_title("Polar Stereographic Grid")
+plt.tight_layout()
+plt.show()
```
jessepisel, line 78: The only thing that I could see differently is adding a plot of the data for creating the geographic grid earlier on in the example. But the more I think about it, the more I like it down here where I can visually compare the geographic and projected coordinates. I think it looks great the way it is.

examples/project_grid.py
leouieda: Thanks! I did mostly because I had to dig into the GMT code in order to see what they did. It's fancier than this but Verde is much easier to install :)

examples/project_grid.py
leouieda: That would be better in a tutorial. But since the plots in the gallery appear all at the top, it wouldn't make much difference. It took me a while to think of a way to show this function in action (we don't have any grids in Verde as sample data).

verde/projections.py
```diff
@@ -0,0 +1,161 @@
...
+    method : string or Verde gridder
+        If a string, will use it to create a :class:`~verde.ScipyGridder` with
+        the corresponding method (nearest, linear, or cubic). Otherwise, should
+        be a gridder/estimator object, like :class:`~verde.Spline`. Default is
+        ``"linear"``.
+    antialias : bool
+        If True, will run a :class:`~verde.BlockReduce` with a mean function to
+        avoid aliasing when the projection results in oversampling of the data
+        in some areas (for example, in polar projections). If False, will not
+        run the blocked mean.
```
jessepisel, line 104: This is great to have antialiasing in here

verde/projections.py
```diff
@@ -0,0 +1,161 @@
...
+                len(grid.dims)
+            )
+        )
+
+    # Can be set to None for some data arrays depending on how they are created
+    # so we can't just rely on the default value for getattr.
+    name = getattr(grid, "name", None)
+    if name is None:
+        name = "scalars"
+
```
jessepisel, line 130: Data checks look good and print useful and meaningful error messages

verde/projections.py
leouieda: I agree :slightly_smiling_face:

leouieda:
> I assume is because it's straightforward enough to not warrant testing

You'd be surprised! These little functions to get the coordinates right were some of the trickiest ones in the whole project. The tests for that function are the examples in the docstring (doctests). I like those if the function outputs are simple enough that they can be checked against expected results. Then it serves the dual purpose of documentation + test. 

Thanks for the review! Merging this in so it can be part of 1.4.0.

PR #255, opened 2020-04-13
Add optional argument to least_squares to copy Jacobian matrix

### LINKED ISSUE #252: Add optional argument to avoid overwriting Jacobian in least square

The `verde.base.least_squares` solves a linear problem with Tikhonov zero order regularization by rescaling the Jacobian matrix to unit variance columns, so the damping parameter is independent of the units of the misfit. While scaling the Jacobian, a `copy=False` argument is passed to `StandardScaler` so we can save memory by scaling the Jacobian in place. This has the advantage of saving memory, which is a good feature to have for most cases. But on some examples we may want to reuse the computed Jacobian, so we want to avoid to overwrite it.

Would be nice to have an optional `copy` argument on `verde.base.least_squares` that is passed to the `StandardScaler` constructor.

If we decide not  to add this option, I would suggest to add a warning on the docstring explicitly saying that the Jacobian matrix will be overwritten. Also to change the following line:
https://github.com/fatiando/verde/blob/9884a274fc7cc49e213345d21f17d960082eaef0/verde/base/least_squares.py#L48

to:
```python
scaler.fit_transform(jacobian)
```
because `fit_transform` scales the matrix it in place if `copy` is `False`.
This way we can also figure out that the matrix will be overwritten by just reading the code.

santisoler:
Add new `copy_jacobian` optional argument to `least_squares` to avoid scaling Jacobian matrix inplace. Add warning on its docstring warning about memory consumption when `copy_jacobian` is True. Add test function for this new feature.

Fixes #252 

**Reminders**:

### DIFF
--- a/verde/base/least_squares.py
+++ b/verde/base/least_squares.py
@@ -7,7 +7,7 @@ from sklearn.preprocessing import StandardScaler
 from sklearn.linear_model import LinearRegression, Ridge
 
 
-def least_squares(jacobian, data, weights, damping=None):
+def least_squares(jacobian, data, weights, damping=None, copy_jacobian=False):
     """
     Solve a weighted least-squares problem with optional damping regularization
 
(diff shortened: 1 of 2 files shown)


verde/base/least_squares.py
```diff
@@ -17,6 +17,12 @@ def least_squares(jacobian, data, weights, damping=None):
     required for predictions. Doesn't normalize the column means because that
     operation can't be undone.
 
+    .. warning::
+
+        Setting `copy_jacobian` to True will copy the Jacobian matrix, doubling
+        the memory required. Use it only if Jacobian matrix is needed
```
leouieda, line 23:
```suggestion
        the memory required. Use it only if the Jacobian matrix is needed
```

PR #258, opened 2020-05-01
Minor typo fix in README

rowanc1:
Following the instructions of "equally" important contributions .... maybe this is just a ploy to get contributors? :)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -130,7 +130,7 @@ What could you possibly offer?
 We assure you that the little voice in your head is wrong.
 
 **Being a contributor doesn't just mean writing code**.
-Equality important contributions include:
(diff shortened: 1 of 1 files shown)


leouieda:
> maybe this is just a ploy to get contributors? :)

I'm no where near that level of cleverness.

Thanks, Rowan!

leouieda:
This is likely present in all repos in the `fatiando` organization so if you want to get a commit into every single one, now is your chance :slightly_smiling_face:

rowanc1:
Ok! Couldn't resist all those green squares I would get on my profile. Really feels like you should be using version control or something ;) ha, weird.

PR #265, opened 2020-05-29
Add extra_coords on BaseGridder generated objects

santisoler:
After predicting through the `profile`, `scatter` and `grid` methods,
`BaseGridder` returns generated objects like xr.Dataset or pd.DataFrame.
Any extra coordinate passed through the `extra_coords` keyword argument
is now included on those objects as new columns or non-dimension
coordinates (depending on the generated object).

**Reminders**:


verde/base/base_classes.py
```diff
@@ -559,3 +571,30 @@ def get_instance_region(instance, region):
             raise ValueError("No default region found. Argument must be supplied.")
         region = getattr(instance, "region_")
     return region
+
+
+def get_extra_coords_names(coordinates, basename):
```
leouieda, line 576: I would make this a method (maybe private for now) so you don't have to pass in `self.extra_coords_name` and it can be modified by derived classes.

verde/base/base_classes.py
```diff
@@ -87,6 +87,10 @@ class BaseGridder(BaseEstimator):
     # (pd.DataFrame, xr.Dataset, etc)
     dims = ("northing", "easting")
 
+    # The default extra_coords names for generated outputs
+    # (pd.DataFrame, xr.Dataset, etc)
```
leouieda, line 91:
```suggestion
    # The default name for any extra coordinates given to methods below
    # through the `extra_coords` keyword argument. Coordinates are 
    # included in the outputs (pandas.DataFrame or xarray.Dataset) 
    # using this name as a basis.
```

verde/base/base_classes.py
```diff
@@ -234,7 +240,7 @@ def grid(
         data_names=None,
         projection=None,
         **kwargs
-    ):
+    ):  # pylint: disable=too-many-locals
```
santisoler, line 243: @leouieda Do you agree with adding this exception?

leouieda: Yes but I would like to try not making a habit of this. I've been frustrated with this as well but I like how it forces me to really think about the function.

But in this case, this is fine. We're planning refactoring of `grid` soon anyways and can remove that when we do it.

PR #272, opened 2020-06-04
Ensure that cache folder exists before locate test

leouieda:
Since Pooch v1.1.1 the cache folder is only created when a `fetch` is
run. Our test was checking that the cache folder existed before doing
any downloads. So first make a download and then check.

**Reminders**:


leouieda:
Never mind. Amended the commit for 1.5.0 instead. This will be a minor difference between the tag and the Zenodo archive but I can live with that.

PR #277, opened 2020-07-24
Deprecate the sample datasets

### LINKED ISSUE #276: Deprecate datasets module

As decided on one of the last Community Calls, every Fatiando dataset will be available for download through Rockhound.
After https://github.com/fatiando/rockhound/pull/84 is merged, we should add deprecation warnings for the `verde.datasets` module and deprecate it for Verde `v.2.0.0`. The module should be removed along with the `data` folder. Besides, all gallery examples should download their data from Rockhound, therefore it should be added as an optional dependency.

**Are you willing to help implement and maintain this feature?** Yes, but would leave to anyone that wants to tackle it.

santisoler:
Sample data will be removed in Verde 2.0.0 in favour of using the [Ensaio](https://www.fatiando.org/ensaio) library instead. Add a deprecation warning to all fetching and plotting functions and warnings to the documentation and docstrings.

Fixes #276 

**Reminders**:


santisoler:
After removing these functions, the `dataset` module will only contain the synthetic `CheckBoard` class.
Shouldn't we find a new place for it to live, maybe something like `verde/synthetic` (as Harmonica has)?

santisoler:
We should also update the gallery examples to use Rockhound fetch functions instead the ones in `verde.datasets`

leouieda (approved):
@santisoler I updated this to mention Ensaio instead and will merge it as soon as CI are happy. Thanks for pushing this!

leouieda:
> After removing these functions, the dataset module will only contain the synthetic CheckBoard class.
Shouldn't we find a new place for it to live, maybe something like verde/synthetic (as Harmonica has)?

It probably would be better there. We'd need a new PR adding the `verde/synthetic.py` module with CheckerBoard and import it back in `verde.datasets` with a DeprecationWarning.

PR #282, opened 2020-08-14
Add function to convert grid to Dataset

santisoler:
Add a function that converts `coordinates` and `data` as `np.array`s to
`xarray.Dataset`. This would make handling grids easier, with the option of
efficiently saving them as netCDF files. Make `BaseGridder.grid` method to use
this function instead of building the `xarray.Dataset` by itself. Add new
`check_data_names` function that converts a `str` into a `tuple` with a single
`str`. Make `BaseGridder.grid`, `BaseGridder.scatter` and `BaseGridder.profile`
able to take a single `str` for the `data_names` argument. Add test function
for default `data_names`. Change `data_names` on all examples and tutorials to
a single `str`.

**Reminders**:


verde/utils.py
```diff
@@ -207,6 +209,116 @@ def maxabs(*args, nan=True):
     return npmax(absolute)
 
 
+def build_grid(
+    coordinates,
+    data,
+    data_name,
+    dims=("northing", "easting"),
+    extra_coords_name=None,
+    attrs=None,
```
leouieda, line 218: Maybe remove this for now and we can be smarter about adding metadata to the Dataset later on. It's always easier to add later than to remove.

verde/__init__.py
```diff
@@ -15,7 +15,7 @@
     longitude_continuity,
 )
 from .mask import distance_mask, convexhull_mask
-from .utils import variance_to_weights, maxabs, grid_to_table
+from .utils import variance_to_weights, maxabs, grid_to_table, build_grid
```
leouieda, line 18: How about `make_xarray_grid`? Or something like that?

santisoler:
Remove the `attrs` parameter. Users can add those later on the `xr.Dataset`. We might want to add metadata `attrs` that could be used by xarray and GMT when plotting, like units.

verde/utils.py
```diff
@@ -207,6 +209,116 @@ def maxabs(*args, nan=True):
...
+        String or list containing strings with the names of the data arrays.
+    dims : list or None
+        The names of the northing and easting data dimensions, respectively,
+        in the output grid. Must be defined in the following order: northing
+        dimension, easting dimension.
+        **NOTE: This is an exception to the "easting" then
+        "northing" pattern but is required for compatibility with xarray.**
+    extra_coords_name : str or list
+        Name or list of names for extra coordinates. Ignored if coordinates has
+        only two elements.
```
leouieda, line 244: Need to be very clear about what are dims and how they are different from coordinates and extra_coords_name.

santisoler:
@leouieda I tried to expalin a little bit better the difference between dims and extra_coords_name, although it might need more work. I added another example that uses extra_coords. Let me know what do you think about this.

verde/utils.py
```diff
@@ -207,6 +209,141 @@ def maxabs(*args, nan=True):
...
 
+def make_xarray_grid(
+    coordinates,
+    data,
+    data_name,
+    dims=("northing", "easting"),
+    extra_coords_name=None,
+):
+    """
+    Create a :class:`xarray.Dataset` out of gridded data
```
leouieda, line 220:
```suggestion
    Create an :class:`xarray.Dataset` grid from numpy arrays
```

verde/utils.py
```diff
@@ -207,6 +209,141 @@ def maxabs(*args, nan=True):
...
+
+    The functions creates a :class:`xarray.Dataset` out of 2d gridded data
+    including easting and northing coordinates, any extra coordinates (like
+    upward) and data arrays.
```
leouieda, line 224:
```suggestion
    This functions creates an :class:`xarray.Dataset` out of 2d gridded data
    including easting and northing coordinates, any extra coordinates (like
    upward elevation, time, etc) and data arrays.
```

verde/utils.py
```diff
@@ -207,6 +209,141 @@ def maxabs(*args, nan=True):
...
+
```
leouieda, line 225:
```suggestion

    Use this to transform the outputs of :func:`verde.grid_coordinates` and
    the ``predict`` method of a gridder into an :class:`xarray.Dataset`. 
    
    .. note::
    
        This is a utility function to help create 2D grids (i.e., grids with two ``dims`` 
        coordinates). For arbitrary N-dimensional arrays, use :mod:`xarray` directly.

```

verde/utils.py
```diff
@@ -207,6 +209,141 @@ def maxabs(*args, nan=True):
...
+        in the output grid. Must be defined in the following order: northing
+        dimension, easting dimension.
+        **NOTE: This is an exception to the "easting" then
+        "northing" pattern but is required for compatibility with xarray.**
+        The easting and northing coordinates in the :class:`xarray.Dataset`
+        will have the same names as the passed dimensions.
+    extra_coords_name : str or list
+        Name or list of names for any additional coordinates besides the
+        easting and northing ones. Ignored if coordinates has
+        only two elements.
```
leouieda, line 251:
```suggestion
        only two elements. The extra coordinates are non-index coordinates of
        the grid array.
```

verde/utils.py
```diff
@@ -207,6 +209,141 @@ def maxabs(*args, nan=True):
...
+        contain values for a dimension in the order: easting, northing,
+        vertical, etc. All arrays must be 2d and need to have the same *shape*.
+        These coordinates can be generated through
+        :func:`verde.grid_coordinates`.
+    data : array or tuple of arrays
+        Array or tuple of arrays with data values on each point in the grid.
+        Each array must contain values for a dimension in the same order as
+        the coordinates. All arrays need to have the same *shape*.
+    data_name : str or list
+        String or list containing strings with the names of the data arrays.
```
leouieda, line 239: This is inconsistent with `grid` which takes `data_names` as a list always. Should we keep that behaviour to be consistent?

verde/utils.py
```diff
@@ -207,6 +209,141 @@ def maxabs(*args, nan=True):
...
+        upward    (northing, easting) float64 5.0 5.0 5.0 5.0 5.0 5.0
+    Data variables:
+        dummy     (northing, easting) float64 1.0 1.0 1.0 1.0 1.0 1.0
+
+    """
+    coordinates = check_coordinates(coordinates)
+    data = check_data(data)
+    if isinstance(data_name, str):
+        data_name = (data_name,)
+    # Create coordinates dictionary (with dims)
```
leouieda, line 303:
```suggestion
    # dims is like shape with order (rows, cols) for the array
    # so the first element is northing and second is easting
```

verde/utils.py
```diff
@@ -207,6 +209,141 @@ def maxabs(*args, nan=True):
...
+    coords = {dims[1]: coordinates[0][0, :], dims[0]: coordinates[1][:, 0]}
+    # Add any extra_coords to the dictionary
```
leouieda, line 305:
```suggestion
    # Extra coordinates are handled like 2D data arrays with 
    # the same dims and the data.
```

leouieda (commented):
Left some comments and suggestions but this is looking good! Good to go once these have been addressed.

verde/utils.py
santisoler: Yes, you're right! Firstly I designed it thinking to let it be also a `str`, but then I decided to keep the same behaviour as the `data_name` parameter of the `grid` method, just forgot to change the docs.

santisoler: I'm not entirely sure. I would use this function to create Dataset with a single data array most of the times, so it's a little bit annoying to remember to add the list brackets just to specify the `data_name`. In fact, the parameter name `data_name` doesn't suggest that I should pass several names. So, what if we extend the functionality to accept a single string?

santisoler:
Thanks @leouieda for the review! I left a response to one of your questions. Let me know what do you think.

I still need to make `BaseGridder.grid` method to use this function. Should we do it here or on a separate PR?

santisoler:
@leouieda If you have a spare time, please take a look at this and let me know if you would change something.
I hope I didn't make any backward breaking change. :crossed_fingers:

verde/base/base_classes.py
```diff
@@ -469,7 +476,7 @@ def scatter(
             following order: northing dimension, easting dimension.
             **NOTE: This is an exception to the "easting" then
             "northing" pattern but is required for compatibility with xarray.**
-        data_names : list of None
+        data_names : str, list or None
             The name(s) of the data variables in the output dataframe. Defaults
             to ``['scalars']`` for scalar data,
```
santisoler, line 481:
```suggestion
            to ``'scalars'`` for scalar data,
```

verde/base/base_classes.py
```diff
@@ -572,7 +580,7 @@ def profile(
...
```
santisoler, line 585:
```suggestion
            to ``'scalars'`` for scalar data,
```

verde/base/utils.py
```diff
@@ -104,6 +104,28 @@ def check_data(data):
...
+    the ``data_names`` argument.
+
+    Examples
+    --------
+
+    >>> check_data_names("dummy")
+    ('dummy',)
+    >>> check_data_names(("component_x", "component_y"))
+    ('component_x', 'component_y')
+    """
```
santisoler, line 123:
```suggestion
    >>> check_data_names(["dummy"])
    ['dummy']
    """
```

PR #293, opened 2020-10-23
Prevent globally overriding warnings.simplefilter

### LINKED ISSUE #290: Change the scope of warning settings to avoid import time side effects
Hi team! Super cool project here. We're excited to be using it, and so far it's working great.

One thing we noticed is that the `warnings` filter set in `vector.py` and `spline.py` and elsewhere, where you call

```
warnings.simplefilter("default")
```

is specified such that simply by calling `import verde` anywhere else overrides existing warning filters with this particular, extremely broad setting.

As a simple example, in a fresh Python 3.7.6 shell I can call

```
import pickle
pickle.dump('test', open('test.pkl', 'wb'))
```

and I see no warning, but if I call

```
import verde
import pickle
pickle.dump('test', open('test.pkl', 'wb'))
```

I see

```
/usr/local/bin/ipython:1: ResourceWarning: unclosed file <_io.BufferedWriter name='test.pkl'>
  #!/usr/local/opt/python/bin/python3.7
ResourceWarning: Enable tracemalloc to get the object allocation traceback
```

In practice, this means we need to manually change a bunch of warning settings in a downstream package simply because we are importing `verde`!

Any chance you might be open to placing your warning settings inside a context manager? Something like this:

```
import warnings

with warnings.catch_warnings():
    warnings.simplefilter("default")
    <any code that triggers the warnings you want to hide>
```

That would prevent the settings in `verde` from affecting other packages. If you're interested in going this route, happy to help clarify how to do it.

To see this behavior I'm running Python 3.7.6 via homebrew, OS 10.5.5, and `verde` 1.5.0.

santisoler:
Modify warnings.simplefilter only when needed (raising
DeprecationWarnings) instead of doing it globally affecting every
other Python library.

Fixes #290 

**Reminders**:


santisoler:
@freeman-lab would you review this PR and check if it actually solves the issue.
It works for me at least, but would be nice to hear about your experience.

leouieda (commented):
Looks good to me. I’ll check locally as well when I get a chance.

leouieda:
It seems to work fine for me on my machine.

leouieda:
Apparently there is a good reason for DeprecationWarnings to not show by default: they are meant for developers not end users. We should actually be issuing FutureWarnings instead (which i think do get show by default). So we could actually remove the filter altogether and change our warning type instead. What do you think?

santisoler:
> Apparently there is a good reason for DeprecationWarnings to not show by default: they are meant for developers not end users. We should actually be issuing FutureWarnings instead (which i think do get show by default). So we could actually remove the filter altogether and change our warning type instead. What do you think?

I agree! I've opened #305 to replace every `DeprecationWarning`s for `FutureWarning`s and remove the filter.
I leave this link for future reference about this topic: https://stackoverflow.com/questions/55377831/difference-between-deprecationwarning-pendingdeprecationwarning-and-futurewarni

I'm closing this PR on behalf of #305. Thanks @leouieda for the insight  and @freeman-lab for the issue.

PR #295, opened 2020-10-27
Refactor get_data_names and check functions

santisoler:
Refactor `get_data_names` as a method of `BaseGridder`: simplify
`get_data_names` logic, make `get_data_names` to use `check_data_names` to
convert single string to a tuple and check that `data_names` has the same
number of elements as data. Add `data_names_defaults` as a class attribute of
`BaseGridder`. Move `check_extra_coords_names` to `verde/base/utils.py` along
with the other check functions. Make `check_data_names` to raise error if
`data_names` is `None`. But `BaseGridder.get_data_names` will return default
values if `data_names` is `None`. Rename test functions for `make_xarray_grid`.

**Reminders**:

### DIFF
--- a/verde/tests/test_utils.py
+++ b/verde/tests/test_utils.py
@@ -98,7 +98,7 @@ def test_partition_by_sum_fails_no_partitions():
     assert "Could not find partition points" in str(error)
 
 
-def test_build_grid():
+def test_make_xarray_grid():
     """
     Check if xarray.Dataset is correctly created
     """
@@ -121,7 +121,7 @@ def test_build_grid():
     assert grid.dummy.shape == (5, 6)
 
 
-def test_build_grid_multiple_data():
+def test_make_xarray_grid_multiple_data():
     """
     Check if xarray.Dataset with multiple data is correctly created
     """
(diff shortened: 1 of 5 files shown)


verde/tests/test_utils.py
```diff
@@ -174,13 +174,16 @@ def test_build_grid_invalid_names():
     data = np.ones_like(coordinates[0])
     with pytest.raises(ValueError):
         make_xarray_grid(coordinates, data, data_names=["bla_1", "bla_2"])
+    # data_names equal to None
+    with pytest.raises(ValueError):
+        make_xarray_grid(coordinates, data, data_names=None)
```
leouieda, line 179: Might want to change the docstring here since `data_names` can't be `None`: https://github.com/fatiando/verde/pull/295/files#diff-948da2a0e34f8cfb6c9d6b6498ec131890d1e637c8128eed5629fb5ce854dd7bR252

leouieda (approved):
Minor comment but otherwise this is really great! Merge away

verde/tests/test_utils.py
santisoler: Nice catch! Even `dims` cannot be None!

PR #298, opened 2020-11-06
Update GMT link in Baja bathymetry example

Esteban82:
I update the GMT link.

Is that from the example gallery 15? Maybe this could clarified.


leouieda:
@Esteban82 thanks for the fix! This is used in several places in the GMT tutorial and gallery so not exclusively in "example 15". Plus, the number could change as we update the GMT docs. I included a mention that this is used the GMT gallery and tutorials.

leouieda (approved):
Merging it in as soon as the CI finishes running 👍🏽

PR #308, opened 2021-03-08
Add license notice to every source file

santisoler:
Add a license notice to every source file and a script for automatically
check if the license is present on every file and add it if missing.

Related to fatiando/maintenance#10 and fatiando/maintenance#12

**Reminders**:


leouieda (commented):
Made a small change to add the header part. This is really cool and should probably be its own package actually!

santisoler:
> Made a small change to add the header part. This is really cool and should probably be its own package actually!

Thanks @leouieda ! We can create a package for it in the future. Maybe it could get the header information from a dotfile instead of having it hardcoded.

PR #285, opened 2020-09-15
WIP Add predict_onto_grid method to BaseGridder

santisoler:
Add new method to `BaseGridder` to predict into regular grid points passed as a `xr.Dataset`.
The predicted values are added to the passed `xr.Dataset` as a new data array.
Add example and test functions.

**Reminders**:

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst` and the base `__init__.py` file for the package.
- [ ] Write detailed docstrings for all functions/classes/methods. It often helps to design better code if you write the docstrings first.
- [ ] If adding new functionality, add an example to the docstring, gallery, and/or tutorials.
- [ ] Add your full name, affiliation, and [ORCID](https://orcid.org) (optional) to the `AUTHORS.md` file (if you haven't already) in case you'd like to be listed as an author on the [Zenodo](https://zenodo.org/communities/fatiando) archive of the next release.

PR #326, opened 2021-05-19
Add option to pass coordinates to the grid method

santisoler:
Allow the `grid` method of `BaseGridder` to take the coordinates of the
preexisting grid either as 2d arrays (meshgrid) or as 1d arrays. Raise
a `FutureWarning` if the `spacing`, `shape` or `region` arguments are passed,
since they will be removed in Verde `v2.0.0`. Write a new `meshgrid_from_1d`
function that creates a meshgrid out of 1d horizontal arrays, but also checks
if they have a single dimensions and supports for passing extra coordinates as
2d-arrays.

This PR is an evolution of #285, so #285 should be closed if this PR is merged.

**Reminders**:


verde/base/base_classes.py
```diff
@@ -420,16 +441,39 @@ def grid(
...
+            )
+        # Get grid coordinates from coordinates parameter
+        if coordinates is not None:
+            ndim = check_ndim_easting_northing(*coordinates[:2])
+            if ndim == 1:
+                # Build a meshgrid if easting and northing are 1d
+                coordinates = meshgrid_from_1d(coordinates)
+            else:
+                check_meshgrid(coordinates)
+        # Build the grid coordinates though vd.grid_coordinates
```
santisoler, line 462:
```suggestion
        # Build the grid coordinates through vd.grid_coordinates
```

examples/predict_on_grid_coordinates.py
```diff
@@ -0,0 +1,38 @@
...
+import verde as vd
+
+# Define a CheckerBoard gridder
+region = (0, 5000, -5000, 0)
+gridder = vd.datasets.CheckerBoard(region=region)
+
+# Define a set of grid coordinates as 1d arrays
+easting = np.linspace(*region[:2], 50)
+northing = np.linspace(*region[2:], 50)
+coordinates = (easting, northing)
```
leouieda, line 30: The example is supposed to show how to grid using an existing grid. Maybe best to make a grid and random points with the checkerboard, fit a spline to the points, and predict on the reference grid. That’s what we would use this feature for, right?

examples/predict_on_grid_coordinates.py
```diff
@@ -0,0 +1,38 @@
...
+
+# Use the .grid() method to predict on the grid coordinates
+grid = gridder.grid(coordinates)
```
leouieda, line 33: Using `coordinates` for the grid coordinates is going to be confusing. Probably best to use `coordinates` for data coordinates and `grid_coordinates` for grid coordinates throughout the code and examples.

verde/base/base_classes.py
```diff
@@ -353,40 +358,55 @@ def score(self, coordinates, data, weights=None):
 
     def grid(
         self,
+        coordinates=None,
```
leouieda, line 361: Put new arguments at the end so we don’t break code that doesn’t use keyword arguments. We should set keyword-only arguments from 2.0.0 to avoid this.

verde/base/base_classes.py
```diff
@@ -353,40 +358,55 @@ def score(self, coordinates, data, weights=None):
...
+          :func:`verde.grid_coordinates` can be passed as extra keyword
+          arguments (``kwargs``) to this method.
 
         Use the *dims* and *data_names* arguments to set custom names for the
         dimensions and the data field(s) in the output :class:`xarray.Dataset`.
         Default names will be provided if none are given.
 
         Parameters
         ----------
+        coordinates : tuple of arrays
```
leouieda, line 390:
```suggestion
        grid_coordinates : tuple of arrays
```

verde/utils.py
```diff
@@ -418,6 +412,78 @@ def meshgrid_to_1d(coordinates):
...
+        )
+    easting, northing = np.meshgrid(coordinates[0], coordinates[1])
+    coordinates = (easting, northing, *coordinates[2:])
+    coordinates = check_coordinates(coordinates)
+    return coordinates
+
+
+def check_ndim_easting_northing(easting, northing):
+    """
+    Check if easting and northing coordinates have the same dimensions
```
leouieda, line 475: Flesh out this docstring a bit since the function also returns something.

verde/utils.py
```diff
@@ -418,6 +412,78 @@ def meshgrid_to_1d(coordinates):
...
+    """
+    ndim = np.ndim(easting)
+    if ndim != np.ndim(northing):
+        raise ValueError(
+            "Horizontal coordinates dimensions mismatch. "
+            + f"The easting coordinate has {easting.ndim} dimensions "
```
leouieda, line 481:
```suggestion
            + f"The easting coordinate array has {easting.ndim} dimensions "
```

leouieda (changes requested):
Left a few suggestions but nothing major. Really good work, as always 🙂

verde/base/base_classes.py
santisoler: You're right! Moving them to the bottom.

examples/predict_on_grid_coordinates.py
santisoler: I agreed at first, but then I found that we already use `grid_coordinates` in Verde for the function that creates the coords of a regular grid. Should we use a different name for the coordinates argument?

examples/predict_on_grid_coordinates.py
santisoler: As far as I remember, we agreed that the `grid` method will not handle the creation of the grid coordinates, just predict on the passed coords and handle the projection if any. In this way, in the future we would predict on a regular grid in two steps:
1. Create the grid coordinates with `grid_coordinates`
2. Use the `grid` method to predict on.

Something like:
```
longitude, latitude = vd.grid_coordinates(...)
grid = gridder.grid((longitude, latitude), projection=projection)
```

In this sense this is example is trying to show that behaviour. Does it make sense or should we change it?

leouieda: Yes, but maybe not at this point since we haven't made that change (and won't until 2.0). Right?

santisoler: Absolutely. What I'm asking is that this example should show this feature. On 2.0 we are going to deprecate the `spacing`, `shape`, etc arguments, but we can start instructing people to pass the grid coordinates to the `grid` method since 1.5. Right?

examples/predict_on_grid_coordinates.py
leouieda: Yeah, you're right. `grid_coords` is reasonable I guess

examples/predict_on_grid_coordinates.py
leouieda: Ah right, I guess so. We'd need one or more PRs making this consistent in the docs. But this example is called "Gridding on a pre-existing grid" so it's not showcasing the future syntax.

verde/utils.py
leouieda: 👍🏽  looks good!

examples/predict_on_grid_coordinates.py
santisoler: You're right. So maybe we would need to change the pitch of the example, or maybe just remove it and update the docs on a different PR. What do you think?

examples/predict_on_grid_coordinates.py
santisoler: I'm having second thoughts on this. Is really an issue of having `coordinates` argument on methods like `predict` and in `grid` while the requirements are slightly different? I think having a common `coordinates` arguments for methods and functions in Verde is a feature and not an issue:
- I can autocomplete on every call by just typing `coor<TAB>` (with `grid_coords` it wouldn't work, I would need to delete and start typing `gri`).
- On some places it doesn't mind if the passed coordinates are scattered or regular gridded (forward modelling in Harmonica for instance). So those cases would be an exception where `coordinates` should be used, although grid coordinates could also be passed.
- On the specific `BaseGridder` we would have two methods that do almost the same but for different kind of coordinates: `predict` for scattered coordiantes, `grid` for grid coordinates. I think that difference is sufficient for differentiating the distribution of the coords.

One last argument, if we use `grid_coords`, an simple example of the `grid` method would look like:
```
grid = gridder.grid(grid_coords=...)
```
I can see myself struggling to read that line out loud when teaching.

What do you think?

leouieda: You're absolutely right. Typing `spline.grid(grid_coords...)` is awkward. So here is a proposal:

1. Function arguments are `coordinates` unless you need to specify 2 sets (like `distance_mask`).
2. For examples: `coordinates` means data coordinates and `grid_coords` means grid coordinates since there will usually be the 2 sets and we'll need to distinguish between them. That way the call could be `spline.grid(coordinates=grid_coords...` or `spline.grid(coordinates=vd.grid_coordinates...)`.

leouieda: Now that I write this, I'm not entire happy with requiring giving coordinates all the time. But I don't know what else to do. I would want to write some examples using this syntax to see how it feels but my feeling is that it will be a bit of a blow to general usability.

examples/predict_on_grid_coordinates.py
leouieda: I'm fine with whichever you prefer. Since this is a still a bit tricky, it would be OK to not have a gallery example for this.

examples/predict_on_grid_coordinates.py
santisoler: I agree with the two points above. Just a reminders for the future: _don't underestimate the old design decisions, there might be a reason why we took that road even if we don't remember it_.

Feel free to try it out on examples, there's no rush for this PR. If we are going to change the default behavior of the base class for Verde gridders, it's good to put a lot of thinking into it.

On the note about passing coordinates every time I think we are already doing it, but hiding it through the `spacing`, `shape`, `region`, `projection` arguments. So the `grid` method currently does two things: creates a set of grid coordinates and predicts on them. I would like to split it in two:
1. Leave the coordinate generation to the user. This way they can corroborate that the grid is the right one before they have to wait until the prediction is carried out. Or they can just use a predefined grid and skip this step.
2. Use the `grid` method to predict on the given coordinates. It should also project the coords before running the prediction if any `projection` is passed. This simplifies the usage of the `grid` method, makes it easier to generalize to gridders like EQL with `easting`, `northing` and `height` (they can be part of the `coordinates` directly) and makes it easier to understand what the `projection` callback will be used for (I always struggle to think if the projection is applied _before_ the prediction or _after_ it and which coordinates should I use for `region` and `spacing`).

Finally, I think this new `grid` method will be _doing a single task, and doing it well_ 😁 .

santisoler:
@leouieda I've added a `FutureWarning` if the `region`, `spacing` or `shape` arguments are passed, in agreement with our last Developer Meeting. Let me know if we need to change something else here. After this PR is merged, we should start changing the current examples so they create the grid coordinates right before calling the grid method.

PR #327, opened 2021-05-20
Use keywords for default arguments in tests and examples

santisoler:
Explicitly pass default arguments with their corresponding keywords on tests
and examples. Particularly for grid_coordinates and BaseGridder.grid calls.
It's better if we instruct people to use keywords of default arguments instead
of passing them by order, this way can add or change the order of default
arguments without breaking backward compatibility.

**Reminders**:


leouieda (approved):
👍🏽 

We should think about making our functions/methods take keyword-only arguments. That would have avoided this type of thing.

PR #328, opened 2021-07-19
Set dims to longitude, latitude if projection passed

santisoler:
Make `BaseGridder.grid`, `BaseGridder.scatter` and `BaseGridder.profile` to use
`("latitude", "longitude")` as default `dims` if a `projection` is passed.

After https://github.com/fatiando/harmonica/issues/233

**Reminders**:


santisoler:
We're breaking some backward compatibility by returning new set of `coord` in the generated outputs, that's why I had to change some tests. Is it OK or should we leave this PR for `2.0.0`?

leouieda (changes requested):
Yep, probably best to wait for 2.0. We should start a branch to gather these PRs.

verde/base/base_classes.py
```diff
@@ -641,12 +644,14 @@ def profile(
...
         return pd.DataFrame(dict(columns), columns=[c[0] for c in columns])
 
-    def _get_dims(self, dims):
+    def _get_dims(self, dims, projection):
         """
         Get default dimension names.
         """
         if dims is not None:
             return dims
+        if projection:
```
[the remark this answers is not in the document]
santisoler: Missed that! Already fixed

verde/base/base_classes.py
```diff
@@ -641,12 +644,14 @@ def profile(
         columns.extend(zip(data_names, data))
         return pd.DataFrame(dict(columns), columns=[c[0] for c in columns])
 
-    def _get_dims(self, dims):
+    def _get_dims(self, dims, projection):
```
leouieda, line 647:
```suggestion
    def _get_dims(self, dims, projection=None):
```

This way classes not passing projection will not break from this change.

leouieda:
I've changed the base here to `2.0-dev` so once the merge conflict is resolved we can merge this.

verde/base/base_classes.py
santisoler: You're right. I'll also change every call of the method so they can pass the projection as a keyword argument (just in case).

PR #329, opened 2021-08-11
Change order of dims in example of make_xarray_grid

santisoler:
Change the order of the dimensions of `DataArrays` in the examples of
`make_xarray_grid`, since `xarray v0.19.0` prints them in a different order.
Tell `pytest` to ignore the outputs of the `print` statements in the
`make_xarray_grid` examples while running doctests.

**Reminders**:

### DIFF
--- a/verde/utils.py
+++ b/verde/utils.py
@@ -292,7 +292,7 @@ def make_xarray_grid(
     >>> grid = make_xarray_grid(coordinates, data, data_names="dummy")
     >>> print(grid)
     <xarray.Dataset>
-    Dimensions:   (easting: 3, northing: 2)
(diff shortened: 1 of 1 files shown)


santisoler:
@leouieda The `xarray` version that is changing the default behavior is `0.19.0`, while on Python3.6 conda installs `xarray` `0.18.2`.
Now we have a single example that prints different outputs depending on the version you are running. How can we tackle this?

santisoler:
Should we tell pytest to skip running the `print` lines? https://docs.pytest.org/en/latest/how-to/doctest.html#skipping-tests

leouieda:
My idea was more "don't use print in doctests" but this is better since the example is easier to read 👍

PR #333, opened 2021-12-06
Remove normalize argument when creating sklearn solvers

santisoler:
The `normalize` argument in the `sklearn`'s `LinearRegression` and `Ridge`
classes has been deprecated since version 1.0.0.

**Reminders**:

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -10,7 +10,7 @@ dependencies:
     - scipy
     - pandas
     - xarray
-    - scikit-learn>=1.0.0
+    - scikit-learn
     - pooch
     - dask!=2021.03.0
     # Optional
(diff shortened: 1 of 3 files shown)


santisoler:
Thanks for the comment.

> :cry: seems like newer versions of dependencies are breaking on Python 3.6. I think it's time we ditch it to be honest. We could have done so a few years back already.

Sound fair. We should do that on a separate PR.

> But I don't think we should require the very latest scikit-learn. See fatiando/community#40

> From the docs of scikit-learn v0.22, the default is normalize=False anyway. I'm not sure why I set it explicitly here. So you could unpin scikit-learn and just remove the normalize argument to fix the error.

I agree. I took a very conservative choice, but now that you mention it, it's better to not pin scikitlearn.

leouieda (approved):

leouieda:
Thanks!

PR #336, opened 2021-12-10
Update GitHub Actions configuration

leouieda:
Adopt some of the newer configuration from our other projects. Use mamba instead of conda to speed up runs. Test only on the Python version end members. Organize the requirements files and separate dependency types. Run the PyPI building and archive checking in PRs but only deploy on pushes and releases. Add the license check to the style checks.

Also updates the setuptools_scm configuration to generate a source file at build time so it's not a run-time dependency anymore. Had to make some changes to make sure the other parts used the correct version information and avoiding circular imports.

**Reminders**:


leouieda:
The failing check is codecov complaining that there is a bit more code in this version 🙂

leouieda:
Merging this in since it's holding things back a bit and it's not breaking anything.

PR #344, opened 2021-12-14
Deprecate the verde.test function

leouieda:
It's not needed and hardly ever used. The same functionality can be achieved with `pytest --pyargs verde` (testing the installed version). Will be removed in v2.0.0.

Related to fatiando/community#44

**Reminders**:


leouieda:
@santisoler we should probably do this for all other packages. It's a bit of a useless function and we don't even use it ourselves.

PR #360, opened 2022-03-18
Update the new 2022 black syntax

leouieda:
**Description**

Black is now out of beta and has introduced a small change in the
syntax, removing white space around `**` operators. They also now
promise to only update syntax like this at most once a year. Seems
reasonable and not a lot of work to keep up to date without pinning
black.

**Relevant issues/PRs**

Reason why #346 is broken. After merging, should work fine.


leouieda:
See https://github.com/psf/black/releases/tag/22.1.0

PR #362, opened 2022-04-21
Replace deprecated numpy dtypes np.bool and np.int

santisoler:
**Description**

Numpy dtypes like `np.bool` and `np.int` have been deprecated since Numpy v1.2.0 in favor of using
builtin `bool` or `int`, or even `np.int64` and `np.int32` which explicitly
sets their precision.

See https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated
for a list of deprecated data types.


leouieda:
Failing because pytest dropped 3.6... Let's get rid of it too and then merge this.

PR #364, opened 2022-04-24
Drop support for Python 3.6

### LINKED ISSUE #342: Drop support for Python 3.6
We could probably have done so a while ago. But it's starting to cause a few problems now (see #339) so it's a good time to do it. Tasks to make this happen:

* Add a warning to the Changelog of v1.7.0 (see #341) that this is the last version to support 3.6
* After 1.7.0 is released, remove Python 3.6 from:
  * `.github/workflows/test.yml` (set the minimum to 3.7)
  * Revert #339 
  * Configuration in `setup.py`
  * Documentation in the Install page

leouieda:
**Description**

Remove the compatibility metadata, remove from the CI matrix, bump the
`python_requires` to 3.7+.

**Relevant issues/PRs**

Fixes #342

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,7 +42,7 @@ jobs:
           - macos
           - windows
         python:
-          - "3.6"
+          - "3.7"
           - "3.10"
         dependencies:
           - latest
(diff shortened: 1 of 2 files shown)


leouieda:
Merging this in to unlock #362. Had previously discussed with @santisoler during the Transform2022 sprints.

PR #380, opened 2022-09-02
Add `scoring` parameter for SplineCV

### LINKED ISSUE #379: Add a 'scoring' argument to SplineCV
**Description of the desired feature:**

As pointed out in https://github.com/orgs/fatiando/discussions/61 by @JamesSample, the `SplineCV` class never got a `scoring` argument when it was added to `cross_val_score`. This means that it's impossible to set a custom scoring function or even prepare the code for the change coming in 2.0 (see #322).

We need to add a `scoring` parameter to `SplineCV` that defaults to `None`. It should be passed along to the call to `cross_val_score` here: https://github.com/fatiando/verde/blob/main/verde/spline.py#L225

The call above that doesn't need to be changed since the `client` argument is going away as well soon. 

A test would be needed to make sure that the `scores_` attribute match a negative MSE instead of R² when the argument is passed.

**Are you willing to help implement and maintain this feature?** Yes but happy for someone else to do this!

JamesSample:
I've added `scoring` as an additional parameter in `SplineCV` to control the scoring metric used during cross-validation. I've also added a simple test to `test_spline.py`, which checks that results from `SplineCV` using the new scoring parameter match those from standard `Spline` combined with `cross_val_score`. I'm not sure I've put this test in the right place - please let me know if it should be elsewhere.

**Relevant issues/PRs:**

(Hopefully) fixes #379 (see also the discussion [here](https://github.com/orgs/fatiando/discussions/61)).

### DIFF
--- a/verde/spline.py
+++ b/verde/spline.py
@@ -93,10 +93,10 @@ class SplineCV(BaseGridder):
         allow mod:`dask` to execute the grid search in parallel (see note
         above).
     scoring : None, str, or callable
-        A scoring function (or name of a function) known to scikit-learn. See
-        the description of *scoring* in
+        The scoring function (or name of a function) used for cross-validation.
+        Must be known to scikit-learn. See the description of *scoring* in
         :func:`sklearn.model_selection.cross_val_score` for details. If None,
-        will fall back to the estimator's ``.score`` method.
+        will fall back to the :meth:`verde.Spline.score` method.
 
     Attributes
     ----------
--- a/verde/spline.py
+++ b/verde/spline.py
@@ -92,6 +92,11 @@ class SplineCV(BaseGridder):
         If True, will use :func:`dask.delayed` to dispatch computations and
         allow mod:`dask` to execute the grid search in parallel (see note
         above).
+    scoring : None, str, or callable
+        A scoring function (or name of a function) known to scikit-learn. See
+        the description of *scoring* in
+        :func:`sklearn.model_selection.cross_val_score` for details. If None,
+        will fall back to the estimator's ``.score`` method.
 
     Attributes
     ----------
@@ -134,6 +139,7 @@ class SplineCV(BaseGridder):
         cv=None,
         client=None,
         delayed=False,
+        scoring=None,
     ):
         super().__init__()
         self.dampings = dampings
@@ -143,6 +149,7 @@ class SplineCV(BaseGridder):
         self.cv = cv
         self.client = client
         self.delayed = delayed
+        self.scoring = scoring
         if engine != "auto":
             warnings.warn(
                 "The 'engine' parameter of 'verde.SplineCV' is "
@@ -229,6 +236,7 @@ class SplineCV(BaseGridder):
                     weights=weights,
                     cv=self.cv,
                     delayed=self.delayed,
+                    scoring=self.scoring,
                 )
                 scores.append(dispatch(np.mean, delayed=self.delayed)(score))
         best = dispatch(np.argmax, delayed=self.delayed)(scores)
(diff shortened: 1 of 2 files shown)


verde/spline.py
```diff
@@ -92,6 +92,11 @@ class SplineCV(BaseGridder):
         If True, will use :func:`dask.delayed` to dispatch computations and
         allow mod:`dask` to execute the grid search in parallel (see note
         above).
+    scoring : None, str, or callable
+        A scoring function (or name of a function) known to scikit-learn. See
+        the description of *scoring* in
+        :func:`sklearn.model_selection.cross_val_score` for details. If None,
+        will fall back to the estimator's ``.score`` method.
```
leouieda, line 99:
```suggestion
        The scoring function (or name of a function) used for cross-validation.
        Must be known to scikit-learn. See the description of *scoring* in
        :func:`sklearn.model_selection.cross_val_score` for details. If None,
        will fall back to the :meth:`verde.Spline.score` method.
```

verde/tests/test_spline.py
```diff
@@ -99,6 +100,25 @@ def test_spline():
...
 
+def test_spline_cv_scoring():
+    "Check scoring parameter works with SplineCV"
+    region = (100, 500, -800, -700)
+    synth = CheckerBoard(region=region)
+    data = synth.scatter(size=1500, random_state=1)
+    coords = (data.easting, data.northing)
+    # Compare SplineCV to results from Spline with cross_val_score
+    for score in ["r2", "neg_root_mean_squared_error"]:
+        spline = Spline()
```
leouieda, line 111:
```suggestion
        spline = Spline(damping=None, mindist=1e-5)
```

Best to set the same parameters here explicitly to avoid breaking in case the default ever changes.

verde/tests/test_spline.py
```diff
@@ -99,6 +100,25 @@ def test_spline():
...
+        score_spline = np.mean(
+            cross_val_score(spline, coords, data.scalars, scoring=score)
+        )
+        # Limit SplineCV to a single parameter set equal to Spline's defaults
+        spline_cv = SplineCV(mindists=[1e-5], dampings=[None], scoring=score)
+        spline_cv.fit(coords, data.scalars)
+        score_spline_cv = spline_cv.scores_.max()
+        npt.assert_allclose(score_spline, score_spline_cv, rtol=1e-5)
```
leouieda, line 119: Any particular reason to check the mean against the max? Would checking all scores individually work for this (like below)?

```suggestion
        score_spline = cross_val_score(spline, coords, data.scalars, scoring=score)
        # Limit SplineCV to a single parameter set equal to Spline's defaults
        spline_cv = SplineCV(mindists=[1e-5], dampings=[None], scoring=score)
        spline_cv.fit(coords, data.scalars)
        npt.assert_allclose(score_spline, spline_cv.scores_, rtol=1e-5)
```

leouieda (commented):
Thank you for adding this! Made a few minor suggestions and a question at the end about the test. Otherwise, this looks very good to me. Just waiting for the CI to finish running tests and checks to see if there is anything else that needs addressing.

verde/tests/test_spline.py
JamesSample: If I understand correctly, the `scores_` parameter of `SplineCV` contains the mean of the 5-fold CV scores for each combination of `mindists` and `dampings` i.e. the number of elements in `spline_cv.scores_` is equal to the number of parameter combinations (which is just one in this example). By contrast, `cross_val_score` returns the score for each fold (a 5-element array in this example).

So I think we do need to compare the mean from `cross_val_score` with the single result from `SplineCV`. However, I agree using `spline_cv.scores_.max()` in this case is confusing, because it's the "max" of a one-element array. Would

    score_spline_cv = spline_cv.scores_[0]

be clearer?

leouieda: Ah right, I forgot that `SplineCV` takes the mean. Yep, your suggestion is better and more explicit about what its doing 👍🏾

leouieda (approved):
Looks great! Merging this in 🚀

leouieda:
@JamesSample please see above ☝🏾 about adding yourself as an author 🙂 Thanks for taking the time to report and implement this!

PR #383, opened 2022-10-21
Simplify the GitHub Actions test workflow matrix

leouieda:
Run on oldest Python and dependencies and newest Python and dependencies instead of every combination of both. This covers the functionality that we expect to support and produces many fewer workflow runs. Update the cache action to v3 as well.

**Relevant issues/PRs:**


leouieda:
Verde doesn't yet implement version bounds on dependencies. This PR needs to wait for that to happen.

PR #384, opened 2022-10-24
Set lower bounds for dependencies based on NEP29

leouieda:
Support all minor versions of Python from the past 42 months and all minor versions of dependencies from the past 24 months. Add a document to the docs explaining the policy. Add jobs to the CI to test oldest and newest versions.

**Relevant issues/PRs:**


leouieda:
TODO:

* Add compatibility to toctree

PR #386, opened 2022-10-24
Use PyGMT instead of Cartopy in Chain tutorial

leouieda:
The PyGMT maps require less code and tend to look nicer than the Cartopy ones. It's also faster for non-rectangular maps. Add PyGMT and GMT as docs dependencies with pinned versions to avoid breaking the docs build. Replacing in a single tutorial to establish the infrastructure. Replace in other places in follow-ups.

**Relevant issues/PRs:**


leouieda:
@santisoler I took over your branch from 2020 and made it so we can do this one page at a time so that the PR doesn't live forever.

leouieda:
Aaand of course there are issues with Cartopy and GDAL. Have to pin matplotlib to avoid issues.

PR #390, opened 2022-10-24
Add a 1D version of grid_coordinates

### LINKED ISSUE #138: Add support for generating N-dimensional grids

Grids are generated by the `BaseGridder.grid` method, which relies on `coordinates.grid_coordinates` to generate the coordinates for prediction. There is a heavy assumption of only 2 dimensions for the grid in the code: `region` is assumed to be `[W, E, S, N]`, assigning coords to `xarray.Dataset` is hard-coded, etc. This is fine for gridders in this package because none of them support predicting data in 3D from 2D observations. 

This is not the case for harmonic functions, like gravity and magnetic data. Because of Green's identities, you can predict these data anywhere in space from observations on a surface. This is know as the equivalent layer technique and it will be implemented in [harmonica](https://github.com/fatiando/harmonica) using Verde as a basis. So it would be good to:

* Add support for `grid_coordinates` to take N-dimensional `region`, `spacing`, and `shape`. For example, `grid_coordinates(region=[0, 10, 0, 20, 0, 30], shape=(10, 11, 12))` should produce three 3D arrays.
* Change `BaseGridder.grid` to set the `Dataset` `coords` attribute dynamically depending on the shape of the coordinates produced by `grid_coordinates`.

leouieda:
Instead of adding the ability to generated 3D coordinates to grid_coordinates, it's more straightforward to add a 1D version of numpy.linspace that can take a spacing instead of a number, knows how to adjust spacing/region, and can return pixel-registered coordinates. This function is then used in grid_coordinates and can be used to generate n-dimensional coordinates if desired (by calling numpy.meshgrid).

**Relevant issues/PRs:**

Fixes #138


leouieda:
Had a hard time naming this. Decided to use `linspace` since it does what we would have liked `numpy.linspace` to do. But using it this way can make it confusing in the code and also in user code (`vd.linspace` and `np.linspace` are very similar). Any suggestion for a better name would be very welcome.

cc @santisoler @LL-Geo @MGomezN @aguspesce

leouieda:
How about `line_coordinates` for lack of a better name?

PR #401, opened 2022-12-09
Use a better Spline Green's function for small distances

leouieda:
If distance is small change the Green's function to `x(log(x**x)-1)` which evaluates to the correct limit of 0. Deprecate the `mindist` parameter in `Spline` and `SplineCV` since it's no longer needed. Remove use of this parameter from the documentation and present `damping` as the only configurable parameter for the splines.


leouieda:
TODO:

- [ ] Implement this on the SplineCV
- [ ] Make sure examples don't set the mindist parameter.
- [ ] Open an issue for removing mindist entirely for v2.0.0.

PR #415, opened 2023-09-22
Remove mention of "setup.py" from pyproject.toml

santisoler:
Since `setup.py` was removed from Verde, the comment in `pyproject.toml` should be updated and remove "setup.py" from it.


leouieda:
Good catch!

PR #418, opened 2024-02-05
Drop support for Python 3.7 and 3.8

leouieda:
Per our policy, it's time to drop support for these.


leouieda:
Tests pass but upload to codecov is failing. Will merge anyway.

PR #435, opened 2024-03-01
Use Burocrata to check/add license notices

leouieda:
Replace our custom script in `tools` with Burocrata. Simplify the Makefile and the Actions checks and add the license notice configuration to pyproject.toml.


leouieda:
Part of https://github.com/fatiando/community/issues/140

PR #436, opened 2024-03-01
Use Trusted Publishers for PyPI deployment

leouieda:
Setup GitHub Actions as a trusted publisher to replace our long-lived API tokens.


leouieda:
Part of https://github.com/fatiando/community/issues/141

PR #443, opened 2024-04-10
Replace setup.cfg for pyproject.toml

santisoler:
Ditch the `setup.cfg` file and use only `pyproject.toml`. Move flake8 configurations to their own `.flake8` file.

**Relevant issues/PRs:**

Related to fatiando/community#143


leouieda:
Forgot to updated Dependente to v0.3 in the CI.

leouieda:
Thanks for doing this!

leouieda:
Maybe we wait for 2.0 to do this since it will remove the `datasets` module so a lot of that configuration will go away.

PR #445, opened 2024-04-10
Ignore flake warning on io module

santisoler:
Ignore the flake8-builtins `A005` error raised because the `verde.io` module is shadowing the builtin `io` module.


santisoler:
I think for Verde v2.0.0 we should aim to making these modules private. That would make this flake8 configuration obsolete.

leouieda:
Yes we should but it wouldn't change this. flake8 complains about `_io.py` as well...

santisoler:
Oh, good to know.

PR #449, opened 2024-04-17
Fix pandas warnings about .ravel() and .apply()

leouieda:
Pandas was issuing FutureWarnings about calling `Series.ravel`, which could be fixed by calling `np.ravel()` instead, and a keyword argument of the groupby `.apply` changing the default to not include the index in the calculations (which is the right thing to do).

### DIFF
--- a/verde/base/least_squares.py
+++ b/verde/base/least_squares.py
@@ -9,6 +9,7 @@ Functions for least-squares fitting with optional regularization.
 """
 from warnings import warn
 
+import numpy as np
 from sklearn.linear_model import LinearRegression, Ridge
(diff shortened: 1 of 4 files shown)


leouieda:
@santisoler the other warning about `aggregate` I couldn't find a way to fix. I think we'll just have to live with it until Pandas makes a new release. Or we pass strings to `BlockReduce` instead of functions but I'm not a huge fan.

leouieda:
So the pandas warning about `apply` won't work on older pandas... 

I think we should leave it as is and let pandas change their default when they want. It's not like we're doing something completely wrong since the tests pass with or without the new argument.,

PR #455, opened 2024-06-17
Move push to codecov to its own job in Actions

leouieda:
Remove the push to codecov step from the `test` job into a new job that depends on the test job. Upload the coverage reports as artifacts after testing, and reuse the artifacts in the new job. Upload all coverage reports in a single push to Codecov to minimize the number of hits.

**Relevant issues/PRs:**

Related to fatiando/community#151


santisoler:
I'm merging this. Test failures will be solved in #461

PR #458, opened 2024-07-01
Improve docstring for `rolling_window`

santisoler:
Improve the docstring of the `verde.rolling_window` function. Fix the wrong description of the `spacing` argument. Add a new example that defines scatter points instead of gridded points.

**Relevant issues/PRs:**

Fix https://github.com/orgs/fatiando/discussions/155


leouieda:
Closing since this was incorporated into [Bordado](https://www.fatiando.org/bordado/).

PR #459, opened 2024-07-01
Fix examples and tests to pass with Numpy 2

santisoler:
Cast Numpy floats and ints to Python floats so doctests can pass using both Numpy <2 and >=2. Update test that was checking floats with regex.


leouieda (approved):
Thanks!

PR #461, opened 2024-08-27
Use pip to install dependencies in test.yml Action

santisoler:
Update the `test.yml` GitHub Action so it makes use of `pip` to install the collected dependencies, instead of having to install Miniforge. Bump `dependente` version to `0.3.0`. Add `dask[distributed]` to `env/requirements-test.txt`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -52,6 +52,14 @@ jobs:
             python: "3.12"
           - dependencies: optional
             python: "3.12"
+          # test on macos-13 (x86) using oldest dependencies and python 3.9
+          - os: macos-13
+            dependencies: oldest
+            python: "3.9"
+        exclude:
+          # don't test on macos-latest (arm64) with oldest dependencies
+          - os: macos-latest
+            dependencies: oldest
     env:
       REQUIREMENTS: env/requirements-build.txt env/requirements-test.txt
       # Used to tag codecov submissions
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -85,6 +85,11 @@ jobs:
       - name: Fetch git tags
         run: git fetch origin 'refs/tags/*:refs/tags/*'
 
+      - name: Echo architechture
+        run: |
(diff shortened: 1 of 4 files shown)


santisoler:
It seems that when running with `macos-13`, the Python setup doesn't work properly: when installing the packages it's using Python 3.12 even if the setup Python action was set to another version. I'll wait if this gets fixed upstream in GitHub.

leouieda:
Mac tests are failing because pip is downloading source distributions for numpy (oldest) and numba (latest) and then failing to compile them. I have idea why it's doing that because wheels are available on PyPI for these versions.

santisoler:
> It seems that when running with macos-13, the Python setup doesn't work properly: when installing the packages it's using Python 3.12 even if the setup Python action was set to another version. I'll wait if this gets fixed upstream in GitHub.

This was fixed, so they are actually running macos 13 on the `macos-13` runner.

I see two things failing now:

* Tests on macos-13 with oldest dependencies. Wheels for Numpy 1.23.5 are only available for macos 10 and 11 in arm (https://pypi.org/project/numpy/1.23.5/#files). That's why it's trying to build it and failing to do so.

* Tests on macos-latest fail due to similar reason: wheels for `numba==0.60.0` are available only for macos 10 and 11 (https://pypi.org/project/numba/0.60.0/#files). But even in `conda-forge`, there are no Numba release for Python 3.13.

santisoler:
In order to solve the issues with tests on oldest versions, maybe we could try updating the minimum required versions (following [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html)). That would mean dropping support for Python 3.9 and supporting Numpy >=1.24 (although by Dec 18 they'll drop 1.24 too).

leouieda:
Alright, I bumped the Python and numpy versions. Let's see if this works now.

leouieda:
Now it's that dance of trying to find compatible versions of everything...

leouieda:
Fixed in #487

PR #473, opened 2024-11-20
Minor fixes to docs

mdtanker:
When reading through the docs I noticed a few small errors which I've  fix here. 
1) The sphinx cross-referencing to the dask delayed function (:func:`dask.delayed`) was not properly working throughout the docs. It looks like the proper reference for it is :func:`dask.delayed.delayed`.  I've updated all these cross-references.

2) In the typehints of the docstrings the use of an oxford comma seems to result in "or" being treated as an object instead of text; see the docstrings for the scoring parameter of [SplineCV](https://www.fatiando.org/verde/dev/api/generated/verde.SplineCV.html#verde.SplineCV) for an example.  I've removed the oxford comma for a few docstrings that had them. 

I built the docs locally and the changes seem to have worked.

### DIFF
--- a/doc/tutorials_src/model_evaluation.py
+++ b/doc/tutorials_src/model_evaluation.py
@@ -210,7 +210,7 @@ print("Delayed k-fold scores:", scores)
 # In this case, the scores haven't actually been computed yet (hence the
 # "delayed" term). Instead, Verde scheduled the operations with Dask. Since we
 # are interested only in the mean score, we can schedule the mean as well using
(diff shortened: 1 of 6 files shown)


santisoler:
Nice catches @mdtanker! I'm merging this :rocket:

santisoler:
**Squash and Merge summary**

Fix cross-references to `dask.delayed.delayed`. Remove Oxford comma before "or" when listing possible types for arguments. It made the "or" word to be treated as a possible type instead of the connector.

Issue #331, opened 2021-11-03
Release v2.0.0


leouieda:
|                   |          Info          |
|:------------------|:-----------------------|
|**Version number** | v2.0.0                |
|**Target date**    | YYYY/MM/DD             |
|**Zenodo DOI**     | 10.5281/zenodo.XXXXXXX |

This issue is to track progress and deprecations for the upcoming 2.0.0 release. We'll be breaking some API compatibilities with this one. See the issues in the [2.0 milestone](https://github.com/fatiando/verde/milestone/5)

To make the breaking changes, we'll be using a `2.0-dev` branch that will be merged into `master` once we're ready for the release.

## Draft a Zenodo archive (to be done by a manager on Zenodo)

- [ ] Go to the Zenodo entry for this project (find the link to the latest Zenodo release on the `README.md` file)
- [ ] Create a "New version" of it.
- [ ] Delete all existing files
- [ ] Copy the reserved DOI to this issue
- [ ] Update release date
- [ ] Update version number (following Semantic Versioning) with a leading `v` (`v1.5.7`)
- [ ] Add as authors any new contributors who have added themselves to `AUTHORS.md`
- [ ] Review author order to make sure it follows the guidelines on our Authorship Guide
- [ ] Save the release draft

## Update the changelog

- [ ] Generate a list of commits between the last release tag and now: `git log HEAD...v1.2.3 > changes.rst`
- [ ] Edit the list to remove any trivial changes (updates by the bot, CI updates, etc).
- [ ] Organize the list into categories (breaking changes, deprecations, bug fixes, new features, maintenance, documentation).
- [ ] Add a list of people who contributed to the release: `git shortlog HEAD...v1.2.3 -sne`
- [ ] Add the release date and Zenodo DOI badge to the top
- [ ] Replace the PR numbers with links: ``sed --in-place "s,#\([0-9]\+\),\`#\1 <https://github.com/fatiando/PROJECT/pull/\1>\`__,g" changes.rst``
- [ ] Copy the changes to `doc/changes.rst`
- [ ] Make a Markdown copy of the changelog: `pandoc -s changes.rst -o changes.md --wrap=none`
- [ ] Add a link to the new release version documentation in `README.rst`
- [ ] Build and serve the docs locally with `make -C doc all serve` to check if the changelog looks well
- [ ] Open a PR to update the changelog
- [ ] Merge the PR

## Make a release

After the changelog PR is merged:

- [ ] Draft a new release on GitHub
[...]

PR #476, opened 2024-11-29
Version 2.0 development

leouieda:
This is a PR to keep track of development in the 2.0 branch. It **won't** be merged as a PR when 2.0 is ready. 

PRs that add things meant for 2.0 should be made against this branch and not main. This includes documentation overhauls and breaking changes. 

The docs for this branch are deployed to https://www.fatiando.org/verde/dev2.0

See #331 and the [2.0 milestone](https://github.com/fatiando/verde/milestone/5)

**DO NOT MERGE THIS PR**. The default of "squash and merge" is no good here. We'll have to merge this branch preserving the history.


santisoler:
Awesome @leouieda! I configured a branch protection for `2.0-dev` that forbids pushing directly to it (a PR must be opened), and requires branches to be updated before merging into `2.0-dev`.

leouieda:
An interesting example of providing lidar data to make a grid here: https://github.com/gdslab/3dep_tutorial. Would be worth going over how this could be streamlined through Verde.

leouieda:
Tracking changes through this PR is becoming troublesome. The CI is getting outdated here and I don't anticipate any new minor releases before 2.0. If we do need to make one, it may be easier to make a release branch and cherrypick the commits we want instead of keeping this version sandboxed. I'll resolve the conflicts and then think about merging this into main and keeping development concentrated on 2.0 so we can get this out.

PR #487, opened 2025-04-29
Streamline our GitHub Actions workflows

leouieda:
Use caching of pip and conda as much as possible. Use pip instead of conda for the tests now that the datasets module was removed. Remove unused dependencies sphinx-gallery, cartopy, and pooch. Bump minimum version of Python to 3.10 and update dependencies.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -35,7 +35,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       REQUIREMENTS: env/requirements-build.txt env/requirements-docs.txt
-      PYTHON: "3.12"
+      PYTHON: "3.13"
     steps:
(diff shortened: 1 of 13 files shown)


leouieda:
Failing test for numba and Mac seems related to running a parallel numba function on a `ThreadPoolExecutor` (through dask here). See discussion in https://github.com/numba/numba/issues/9288 and https://github.com/numbagg/numbagg/pull/201

leouieda:
Implementing the solution used by numbaag is out of the question. It would require a whole caching system for numba jit since the functions sort of need to be recompiled when entering the threading context. And apparently this can also lead to some segfaulting on some contexts. 

Instead, I'll implement a `parallel` argument that can turn it off like we did in Harmonica. Then we can add docs telling users about the possible crash on Mac when using pip.

Issue #504, opened 2025-05-08
Allow interpolators to receive multiple components


leouieda:
All of our interpolators except for `VectorSpline2D` can only receive a single data vector. We used the `Vector` interpolator to allow combining multiple interpolators to handle data with multiple components. But we're removing that class (#481). Instead, it would be more natural for all interpolators to allow multiple arrays in the `data` argument and then interpolate them all separately. We'd then need to store the model for each component and predict accordingly.

PR #513, opened 2025-05-14
Remove the Vector composite estimator

### LINKED ISSUE #481: Remove the Vector estimator
Thinking about it now, this is not really necessary. It's inconsistent since BlockReduce can receive multiple components and operate on them separately but Spline and Trend can't. Would be better to remove it and make Spline and Trend accept multiple components in the `data` and then fit and predict multiple splines/trends.

leouieda:
It's no longer necessary since the Chain was removed and all other estimators have support for multiple component data.

**Relevant issues/PRs:** Fixes #481


leouieda:
Wait until #504 is implemented. Then rebase and update the model selection tests to use Trend instead of VectorSpline2D.

PR #514, opened 2025-05-14
Organize input validation functions into a module

leouieda:
These are all of the `check_*` functions that are scattered throughout the code. Also take the opportunity to modernize them and combine some that could be made into a single function. Functions `check_data` and `check_coordinates` were combined into `check_tuple_of_arrays`. Function `check_data_names` and `check_extra_coords_names` were combined into `check_names`. Function `check_fit_input` was given more functionality, like ravelling the input arrays and better checking of compatibility between data and weights. The option to unpack data and weights arguments was removed since we're moving towards having estimators always deal with multiple data components.


leouieda:
New functions are finished and tested but still need to be used throughout the code base.

This is paving the way for #504.

PR #524, opened 2025-11-19
Add `percentile` parameter to `maxabs()` function

### LINKED ISSUE #523: Add `quantile` option to `maxabs`
`vd.maxabs()` is often used to set limits for diverging color scales. I find I commonly don't want the max absolute values, but something like the 99% of 95% percentile of max absolute values, so the color scale is not too washed out.  You can simply multiply the returned `maxabs` value by .9, but this is not really the 90% quantile, but just 90% of the max value. I propose adding an optional `quantile` parameter to `maxabs`, which defaults to 1, but can be any float between 0 and 1, representing a decimal percentile of the max absolute values.  

Other packages, like `xarray`, use a similar `robust` parameter in their plotting functions. We could also have a `robust` boolean parameter (defaults to `False`), with an associated `robust_percentile` with a sensible default like .98.

If others think this would be useful, I'm happy to implement it.

mdtanker:
Allow returning a percentile of the absolute value of the provided values, instead of just the maxium absolute value. For example, using the 95% percentile instead of the maximum will create a colorbar that is less washed out and reduces the effects of outliers in the data. 

If multiple arrays are provided, they need to be concatinated before calculating the quantiles, so it can be significantly slower than just the maximum absolute value determination. 

**Relevant issues/PRs:**

Closes #523

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -114,7 +114,7 @@ jobs:
       # Store the docs as a build artifact so we can deploy it later
       - name: Upload HTML documentation as an artifact
         if: github.event_name == 'release' || github.event_name == 'push'
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v5
         with:
           name: docs-${{ github.sha }}
           path: doc/_build/html
@@ -132,7 +132,7 @@ jobs:
 
       # Fetch the built docs from the "build" job
       - name: Download HTML documentation artifact
-        uses: actions/download-artifact@v5
+        uses: actions/download-artifact@v6
         with:
           name: docs-${{ github.sha }}
           path: doc/_build/html
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -81,7 +81,7 @@ jobs:
       - name: Upload archives as artifacts
         # Only if not a pull request
         if: success() && github.event_name != 'pull_request'
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v5
         with:
           name: pypi-${{ github.sha }}
           path: dist
@@ -108,7 +108,7 @@ jobs:
 
       # Fetch the built archives from the "build" job
       - name: Download built archives artifact
-        uses: actions/download-artifact@v5
+        uses: actions/download-artifact@v6
         with:
           name: pypi-${{ github.sha }}
           path: dist
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -166,7 +166,7 @@ jobs:
         run: coverage xml
 
       - name: Upload coverage report as an artifact
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v5
         with:
           name: coverage_${{ matrix.os }}_${{ matrix.dependencies }}
           path: ./coverage.xml
@@ -185,7 +185,7 @@ jobs:
         # Maximum coverage is achieved by combining reports from every runner.
         # Each coverage file will live in its own folder with the same name as
         # the artifact.
-        uses: actions/download-artifact@v5
+        uses: actions/download-artifact@v6
         with:
           pattern: coverage_*
 


santisoler (changes requested):
Thanks @mdtanker for opening this PR! I left a few minor suggestions below.

It would be nice to add tests for those two errors after invalid type and value of `percentile`, just to have coverage on them. It would also be nice to include a test where we pass multiple arrays to the function, just to check that the concatenation works ok. Optionally, add arrays with different dimensions to support the case I explain in the comment below, but that's not 100% necessary.

Let me know if you need any help with those!

verde/utils.py
```diff
@@ -216,14 +224,38 @@ def maxabs(*args, nan=True):
...
+        else:
+            npmin, npmax = np.min, np.max
+        absolute = [npmax(np.abs([npmin(i), npmax(i)])) for i in arrays]
+        return npmax(absolute)
     else:
-        npmin, npmax = np.min, np.max
-    absolute = [npmax(np.abs([npmin(i), npmax(i)])) for i in arrays]
-    return npmax(absolute)
+        if isinstance(percentile, (int, float)) is False:
+            raise ValueError("Percentile must be a float or an integer.")
```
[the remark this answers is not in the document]
mdtanker: Agreed!

verde/utils.py
```diff
@@ -216,14 +224,38 @@ def maxabs(*args, nan=True):
...
+        if percentile < 0 or percentile > 100:
+            raise ValueError("Percentile must be between 0 and 100.")
+        if nan:
+            nppercentile = np.nanpercentile
+        else:
+            nppercentile = np.percentile
+        combined_array = np.concatenate(arrays)
+        return nppercentile(np.abs(combined_array), percentile)
```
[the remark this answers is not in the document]
mdtanker: Oh good catch, will add a test for this as well.

verde/tests/test_utils.py
```diff
@@ -334,3 +335,55 @@ def test_check_ndim_easting_northing():
...
+    assert (
+        pytest.approx(
+            float(maxabs([0, 1, 2, 3, 4], [[-2, 2], [0, 5]], percentile=80)), 0.1
+        )
+        == 3.4
+    )
+
+    # test invalid percentile types
+    with pytest.raises(TypeError):
+        maxabs(data, percentile="90")
```
santisoler: Adding checks for error messages, so we are sure the errors raised are the ones we added to the function:
```suggestion
    msg = "Invalid 'percentile' of type"
    with pytest.raises(TypeError, match=msg):
        maxabs(data, percentile="90")
```

verde/tests/test_utils.py
```diff
@@ -334,3 +335,55 @@ def test_check_ndim_easting_northing():
...
+    with pytest.raises(TypeError):
+        maxabs(data, percentile=[90])
```
santisoler:
```suggestion
    msg = "Invalid 'percentile' of type"
    with pytest.raises(TypeError, match=msg):
        maxabs(data, percentile=[90])
```

verde/tests/test_utils.py
```diff
@@ -334,3 +335,55 @@ def test_check_ndim_easting_northing():
...
+    with pytest.raises(TypeError):
+        maxabs(data, percentile=None)
```
santisoler:
```suggestion
    msg = "Invalid 'percentile' of type"
    with pytest.raises(TypeError, match=msg):
        maxabs(data, percentile=None)
```

verde/tests/test_utils.py
```diff
@@ -334,3 +335,55 @@ def test_check_ndim_easting_northing():
...
+
+    # test invalid percentile values
+    with pytest.raises(ValueError):
+        maxabs(data, percentile=-10)
+    with pytest.raises(ValueError):
+        maxabs(data, percentile=110)
```
santisoler:
```suggestion
    msg = "Invalid 'percentile' value of"
    with pytest.raises(ValueError, match=msg):
        maxabs(data, percentile=110)
```

mdtanker:
Here are some timings for future reference:

```python
import numpy as np
import verde as vd
a = np.random.uniform(size=(27, 100))
```
------------

* Using `np.nanmin` or `np.nanmax` is ~x2 slower than the non-nan functions

  ```python
  %timeit vd.maxabs(a, nan=True)
  %timeit vd.maxabs(a, nan=False)
  
  >>>40.7 μs ± 4.39 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
  >>>18.6 μs ± 693 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
  ```

* Using percentile is ~3-4x slower than min/max

  ```python
  %timeit vd.maxabs(a, percentile=100, nan=True)
  %timeit vd.maxabs(a, percentile=99, nan=True)
  
  >>>40.1 μs ± 3.13 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
  >>>141 μs ± 12.1 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
  ```

* `np.nanpercentile` is ~5-10x slower than `np.percentile`
 
    ```python
    %timeit vd.maxabs(a, percentile=100, nan=False)
    %timeit vd.maxabs(a, percentile=99, nan=False)
    
    >>>18.9 μs ± 1.02 μs per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
    >>>129 μs ± 10.3 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
    ```

verde/utils.py
```diff
@@ -187,11 +187,19 @@ def maxabs(*args, nan=True):
     args
         One or more arrays. If more than one are given, a single maximum will
         be calculated across all arrays.
+    nan : bool, optional
+        If True, will use the ``nan`` version of numpy functions to ignore
+        NaNs.
+    percentile : float, optional
+        Instead of return the maximum absolute value, return a given
+        percentile and the absolute values. Must be between 0 and 100. A
```
santisoler:
```suggestion
        percentile of the absolute values. Must be between 0 and 100. A
```

Issue #525, opened 2025-11-20
Add a function to get the min and max (or supplied percentiles) of values


mdtanker:
**Description of the desired feature:**
This function would be the equivalent of `maxabs` but to use for datasets not diverging around 0. For example, for plotting topography data, you may want to get the max and min values of the grid to create a colormap. Similar to #524, I propose allowing passing a tuple of 2 percentiles, so users can get more robust values, such as the 2% and 98% percentiles of the data, instead of the min and max, to help colorbars be less sensitive to outliers.

**Are you willing to help implement and maintain this feature?** 

Yes, I will open a PR.

PR #526, opened 2025-11-20
add new `minmax` function

mdtanker:
Adds an equivalanet function to `maxabs` but for calculating the min and max values of arrays, or optionally user-specified percentiles of the values. 

This is useful for 
1) if you want to plot a series of datasets with the same colorscale so you need to calculate the overall min/max for all the datasets
2) if you want to get robust colormap limits, excluding outliers by using percentiles, such as the 2nd and 98th percentiles.

**Relevant issues/PRs:**

Implements the function requested in #525 
Follows the percentiles approach from #524 and #523

### DIFF
--- a/verde/utils.py
+++ b/verde/utils.py
@@ -359,7 +359,7 @@ def minmax(*args, nan=True, min_percentile=0, max_percentile=100):
         npmax = np.max
         nppercentile = np.percentile
 
-    if min_percentile == 0 and max_percentile == 0:
+    if min_percentile == 0 and max_percentile == 100:
         min_ = npmin([npmin(i) for i in arrays])
         max_ = npmax([npmax(i) for i in arrays])
         return min_, max_


verde/utils.py
```diff
@@ -226,6 +226,104 @@ def maxabs(*args, nan=True):
...
+
+    if percentile == (0, 100) or percentile == [0, 100]:
+        if nan:
+            npmin, npmax = np.nanmin, np.nanmax
+        else:
+            npmin, npmax = np.min, np.max
+        # Get min and max of each array
+        mins_maxs = [([npmin(i), npmax(i)]) for i in arrays]
+        # Get min of mins and max of maxs
+        return (npmin([i[0] for i in mins_maxs]), npmax([i[1] for i in mins_maxs]))
```
santisoler: Minor suggestion to make this more explicit:
```suggestion
        min_ = npmin([npmin(i) for i in arrays])
        max_ = npmax([npmax(i) for i in arrays])
        return min_, max_
```

verde/utils.py
```diff
@@ -226,6 +226,104 @@ def maxabs(*args, nan=True):
     return npmax(absolute)
 
 
+def minmax(*args, nan=True, percentile=(0, 100)):
```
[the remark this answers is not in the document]
mdtanker: Yeah that sounds good, I'll add the change.

mdtanker:
Here are some timings for future reference:

```python
import numpy as np
import verde as vd
a = np.random.uniform(size=(27, 100))
```
------------

* Using `np.nanmin` or `np.nanmax` is ~x2 slower than the non-nan functions

  ```python
  %timeit vd.minmax(a, nan=True)
  %timeit vd.minmax(a, nan=False)
  
  >>>48.8 μs ± 1.94 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
  >>>20.2 μs ± 1.84 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
  ```

* Using percentiles is ~4x slower than min/max
    * Only calculating 1 percentile, and using min/max for the other saves some time

  ```python
  %timeit vd.minmax(a, min_percentile=0, max_percentile=100, nan=True)
  %timeit vd.minmax(a, min_percentile=1, max_percentile=99, nan=True)
  %timeit vd.minmax(a, min_percentile=0, max_percentile=99, nan=True)
  
  >>>50.5 μs ± 2.56 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
  >>>289 μs ± 26.5 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
  >>>179 μs ± 3.93 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
  ```

* `np.nanpercentile` is ~5-10x slower than `np.percentile`
 
    ```python
    %timeit vd.minmax(a, min_percentile=0, max_percentile=100, nan=False)
    %timeit vd.minmax(a, min_percentile=1, max_percentile=99, nan=False)
    %timeit vd.minmax(a, min_percentile=0, max_percentile=99, nan=False)
    
    >>>20.7 μs ± 843 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
    >>>232 μs ± 17.7 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
    >>>111 μs ± 5.5 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
    ```

verde/utils.py
```diff
@@ -261,6 +261,137 @@ def maxabs(*args, nan=True, percentile=100):
...
+    Returns
+    -------
+    min, max : float
+        The minimum and maximum (or percentile) values across all arrays.
+
+    Examples
+    --------
+
+    >>> result = minmax((1, -10, 25, 2, 3))
+    >>> tuple(map(float, result))
```
santisoler: We can replace these with a `print` statement that is less cryptic for users than using `map`:
```suggestion
    >>> print(result)
```

verde/utils.py
```diff
@@ -261,6 +261,137 @@ def maxabs(*args, nan=True, percentile=100):
...
+    (-10.0, 25.0)
+    >>> result = minmax(
+    ...     (1, -10.5, 25, 2), (0.1, 100, -500), (-200, -300, -0.1, -499)
+    ... )
+    >>> tuple(map(float, result))
```
santisoler:
```suggestion
    >>> print(result)
```

verde/utils.py
```diff
@@ -261,6 +261,137 @@ def maxabs(*args, nan=True, percentile=100):
...
+
+    If a more robust statistic is desired, you can use ``min_percentile`` and
+    or ``max_percentile`` to get the values at given percentiles instead of
+    the minimum and maximum.
+
+    >>> import numpy as np
+    >>> result = minmax(
+    ...     (1, -10, 25, 2, 3), min_percentile=2, max_percentile=98
+    ... )
+    >>> tuple(map(float, result))
```
santisoler:
```suggestion
    >>> print(result)
```

verde/utils.py
```diff
@@ -261,6 +261,137 @@ def maxabs(*args, nan=True, percentile=100):
...
+    if nan:
+        npmin = np.nanmin
+        npmax = np.nanmax
+        nppercentile = np.nanpercentile
+    elif not nan:
+        npmin = np.min
+        npmax = np.max
+        nppercentile = np.percentile
+    else:
+        raise TypeError(f"The 'nan' parameter ({nan}) must be a boolean.")
```
santisoler: Nit picking here. If we want to check if `nan` is boolean at the same time we assign the functions we'll use later on, I think we should check if `nan is True` or `nan is False`, or else raise the error. With this setup we could pass an object that has the `__bool__` method as `nan` but is not a boolean (not True or False), and the code would still work.

```suggestion
    if nan is True:
        npmin = np.nanmin
        npmax = np.nanmax
        nppercentile = np.nanpercentile
    elif nan is False:
        npmin = np.min
        npmax = np.max
        nppercentile = np.percentile
    else:
        raise TypeError(f"The 'nan' parameter ({nan}) must be a boolean.")
```

verde/utils.py
```diff
@@ -261,6 +261,137 @@ def maxabs(*args, nan=True, percentile=100):
...
+
+        # only calculate min percentile
+        elif min_percentile != 0:
+            min_ = nppercentile(combined_array, min_percentile)
+
+        # only calculate max percentile
+        if max_percentile != 100:
+            max_ = nppercentile(combined_array, max_percentile)
+
+    return min_, max_
```
santisoler: I think we could simplify this logic a bit. What about this:
```suggestion
    if min_percentile == 0 and max_percentile == 0:
        min_ = npmin([npmin(i) for i in arrays])
        max_ = npmax([npmax(i) for i in arrays])
        return min_, max_

    # concatenate values of all arrays
    combined_array = np.concatenate([a.ravel() for a in arrays])
    min_, max_ = nppercentile(combined_array, [min_percentile, max_percentile])
    return min_, max_
```

If any of the `min_percentile` and `max_percentile` is different from the defaults, then we'll use `nppercentile` anyways. And a `min_percentile=0` will always lead to the minimum value. Moreover, I suspect there's no significant computational hit in passing two percentiles, since I guess that Numpy is computing each while looping over the elements in the array once.

So, we would only use `npmin` and `npmax` if `min_percentile==0` **and** `max_percentile==100`.

What do you think?

verde/utils.py
mdtanker: Ok that soudns good to me. I thought `if nan` was exactly equivalent to `if nan is True`.

verde/utils.py
mdtanker: There is a bit of time savings from only calculating one percentile and using min/max for the other. But maybe the time savings is not worth the extra code? At least that's how I interpreted the benchmarking results from below. The first uses min and max, with no percentiles, the second uses 1 and 99 percentiles, and the third uses min and 99 percentiles. As you can see, using 1 percentile and 1 min/max calculation is a bit fast than using 2 percentile calculations. 

What do you think?

```python
%timeit vd.minmax(a, min_percentile=0, max_percentile=100, nan=True)
%timeit vd.minmax(a, min_percentile=1, max_percentile=99, nan=True)
%timeit vd.minmax(a, min_percentile=0, max_percentile=99, nan=True)

>>>50.5 μs ± 2.56 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
>>>289 μs ± 26.5 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
>>>179 μs ± 3.93 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

santisoler: Hmm I would have guessed that passing multiple percentiles would not impact the computation time that much. But nonetheless, those differences are not significant, so I think I would keep the code simpler. Do you agree?

Even for a large array with 100 million elements, the difference is not very noticeable by users:

```python
import numpy as np

a = np.random.default_rng().uniform(size=100_000_000)

# benchmark np.min and np.percentile with a single value
%timeit np.min(a)
%timeit np.percentile(a, 90)
```
```
38.1 ms ± 351 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)
862 ms ± 3.75 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

```python
# benchmark percentile with two values
%timeit np.percentile(a, [0, 90])
```
```
1.01 s ± 5.85 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

verde/utils.py
```diff
@@ -263,6 +263,120 @@ def maxabs(*args, nan=True, percentile=100):
         return nppercentile(combined_array, percentile)
 
 
+def minmax(*args, nan=True, min_percentile=0, max_percentile=100):
+    """
+    Calculate the minimum and maximum values (or percentiles) of the given
+    array(s).
```
leouieda:
```suggestion
    Calculate the minimum and maximum values of the given array(s).
```

It's good to keep the first part of the docstring to a single line since it's what shows up in the API page and a lot of tab completions.

PR #559, opened 2026-08-13
Fix undamped least-squares to keep full precision on scikit-learn >= 1.9 (#558)

### LINKED ISSUE #558: Undamped `Spline` silently return wrong predictions with scikit-learn >= 1.9
**Description of the problem:**

Context: During the bump of scikit-learn in Guix, some Verde tests started failing:
- `test_chain.py::test_chain_double`,
- `test_spline.py::test_spline`,
- all three `test_spline.py::test_spline_cv[...]`

Since scikit-learn 1.9, fitting `verde.Spline` (or `verde.SplineCV` whenever cross-validation selects `damping=None`, which is the common/default outcome for noise-free-ish data) with the default, undamped configuration produces predictions that no longer match the training data. No exception or

This also affects anything that chains an undamped `Spline`, e.g. `Chain([..., ("spline", Spline())])`.

Investigation (disclaimer : partly assisted by generative AI)
`verde.base.least_squares.least_squares()` fits the undamped case via:

```python
regr = LinearRegression(fit_intercept=False)
regr.fit(jacobian, np.ravel(data), sample_weight=weights)
```

Internally, `LinearRegression.fit()`'s dense-data path does:

```python
self.coef_, _, self.rank_, self.singular_ = linalg.lstsq(X, y, cond=self.tol)
```

`self.tol` defaults to `1e-6`. Per scikit-learn's own docstring for
`LinearRegression`:

```
tol : float, default=1e-6
    The precision of the solution (`coef_`) is determined by `tol` which
    specifies the convergence criterion of the underlying solver. ...
    `tol` is set as `cond` of :func:`scipy.linalg.lstsq` when fitting on
    dense training data.

    .. versionadded:: 1.7
    .. versionchanged:: 1.9
       Now supported on dense data, interpreted as the `cond` parameter.
```

**System information**

- verde 1.9.0
- scikit-learn 1.9.0
- python 3.12

youdie006:
Fixes #558.

### Problem

Since scikit-learn 1.9, `LinearRegression.fit` passes its `tol` attribute (default `1e-6`) as the `cond` argument of `scipy.linalg.lstsq`, which truncates small singular values. For the ill-conditioned Jacobians produced by `Spline`, the **undamped** fit (`damping=None`) then silently returns predictions that no longer match the training data. No exception is raised.

This also affects:

- `SplineCV` whenever cross-validation selects `damping=None` (the common outcome for near-noise-free data)
- anything chaining an undamped spline, e.g. `Chain([..., ("spline", Spline())])`

On scikit-learn >= 1.9 this makes `test_spline.py::test_spline`, `test_spline_cv[...]`, and `test_chain.py::test_chain_double` fail.

### Fix

Solve the undamped case directly with `scipy.linalg.lstsq(cond=None)` instead of `sklearn.LinearRegression`, so the full-precision solution is used regardless of the scikit-learn version. Weighted fits are handled by row-scaling the Jacobian and data by `sqrt(weights)`, which is equivalent to `LinearRegression`'s `sample_weight`. The damped path (`Ridge`) is unchanged.

This also moves the undamped path off the scikit-learn estimator, in the direction you mentioned in the issue. I kept the change minimal and surgical (undamped path only) rather than reworking the damped path too; happy to widen the scope if you'd prefer.

### Tests

Red-green verified against scikit-learn 1.9.0:

- The existing real-trigger tests (`test_spline`, `test_spline_cv`, `test_chain_double`) fail on `main` with scikit-learn >= 1.9 and pass with this change.
- Added `test_least_squares_undamped_full_precision`: an ill-conditioned full-rank system whose undamped fit must reproduce the data to near machine precision. Fails before this change (max abs error ~1.8e-7) and passes after (~5e-16).
- Added `test_least_squares_undamped_weights`: uniform weights match the unweighted fit, guarding the new weighted path.

`black`, `isort`, and `flake8` (repo-pinned versions) are clean.

---
Disclosure: I used AI assistance (Claude) while preparing this change. I verified the root cause, ran the tests (red-green), and take responsibility for the contribution.

### DIFF
--- a/verde/base/least_squares.py
+++ b/verde/base/least_squares.py
@@ -10,7 +10,8 @@ Functions for least-squares fitting with optional regularization.
 from warnings import warn
 
 import numpy as np
-from sklearn.linear_model import LinearRegression, Ridge
+from scipy.linalg import lstsq
+from sklearn.linear_model import Ridge
 from sklearn.preprocessing import StandardScaler
 
 
@@ -62,10 +63,29 @@ def least_squares(jacobian, data, weights, damping=None, copy_jacobian=False):
         )
     scaler = StandardScaler(copy=copy_jacobian, with_mean=False, with_std=True)
     jacobian = scaler.fit_transform(jacobian)
+    data = np.ravel(data)
     if damping is None:
-        regr = LinearRegression(fit_intercept=False)
+        # Solve the undamped problem directly with scipy.linalg.lstsq
+        # instead of sklearn's LinearRegression. Since scikit-learn 1.9,
+        # LinearRegression.fit passes its ``tol`` (default 1e-6) as the
+        # ``cond`` argument of scipy.linalg.lstsq, truncating small
+        # singular values. That silently degrades the fit for the
+        # ill-conditioned Jacobians produced by Spline, so undamped
+        # predictions no longer match the training data. Passing
+        # ``cond=None`` keeps the full-precision solution regardless of
+        # the scikit-learn version. See #558.
+        if weights is None:
+            coef = lstsq(jacobian, data, cond=None)[0]
+        else:
+            sqrt_weights = np.sqrt(np.ravel(weights))
+            coef = lstsq(
+                jacobian * sqrt_weights[:, np.newaxis],
+                data * sqrt_weights,
+                cond=None,
+            )[0]
+        params = coef / scaler.scale_
     else:
         regr = Ridge(alpha=damping, fit_intercept=False)
-    regr.fit(jacobian, np.ravel(data), sample_weight=weights)
-    params = regr.coef_ / scaler.scale_
+        regr.fit(jacobian, data, sample_weight=weights)
+        params = regr.coef_ / scaler.scale_
     return params
--- a/verde/tests/test_base.py
+++ b/verde/tests/test_base.py
@@ -467,3 +467,37 @@ def test_least_squares_copy_jacobian():
     npt.assert_allclose(jacobian, original_jacobian)
     least_squares(jacobian, data, weights=None)
     assert not np.allclose(jacobian, original_jacobian)
+
+
+def test_least_squares_undamped_full_precision():
+    """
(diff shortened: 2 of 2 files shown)


youdie006:
Flagging the red `windows python=3.9 dependencies=oldest` job, since I do not think it is caused by this PR and it may need attention independently.

It fails during collection, 59 errors, before any test in the file I touched runs. The traceback is entirely in the import of the installed package and terminates outside Verde:

```
verde/__init__.py:8            from . import datasets, synthetic
verde/datasets/__init__.py:18  from .synthetic import CheckerBoard
verde/datasets/synthetic.py:12 from ..synthetic import CheckerBoard as _CheckerBoard
verde/synthetic.py:10          from .base import BaseGridder
verde/base/base_classes.py:17  from ..coordinates import grid_coordinates, ...
verde/coordinates.py:16        from .utils import kdtree
verde/utils.py:15              import xarray as xr
xarray/__init__.py:1           from . import testing, tutorial, ufuncs
xarray/tutorial.py:13          from .backends.api import open_dataset
xarray/backends/__init__.py:6  from .cfgrib_ import CfGribDataStore
xarray/backends/cfgrib_.py:16  from .locks import SerializableLock, ensure_lock
                               -> from .core import CommClosedError
```

So the oldest-pinned `xarray` reaches into `distributed` at import time and lands on an API that has since moved. This PR changes only `verde/base/least_squares.py` and `verde/tests/test_base.py`, neither of which is on that path.

One caveat on my own claim: I could not confirm the same job is red on `main`, because the `test` workflow has not run there since #557 on 2026-08-04 — the recent green runs on `main` are the Dependabot workflow, not this one. So I can say the failure is not reachable from my diff, but not that it is already known-broken on `main`. It looked worth a note either way.

Happy to leave it alone; pinning the oldest environment is your call and out of scope for a precision fix.

Disclosure: I use AI assistance in my work, and I review and verify everything before it goes out.

